PR #44, opened 2022-01-07
Support prepared statements type inference

staabm:

PR #48, opened 2022-01-08
use proper default return value via `ParametersAcceptorSelector`

staabm:
required for https://github.com/staabm/phpstan-dba/pull/44

relates to https://github.com/phpstan/phpstan-src/pull/916


composer.json
```diff
@@ -4,16 +4,15 @@
 	"require": {
 		"php": "^8.0",
 		"ext-mysqli": "*",
-		"phpstan/phpstan": "^1.2"
+		"phpstan/phpstan": "dev-master"
```
staabm, line 7: because of https://github.com/phpstan/phpstan-src/pull/916

composer.json
```diff
@@ -4,16 +4,15 @@
...
 	},
 	"require-dev": {
 		"ext-pdo": "*",
 		"friendsofphp/php-cs-fixer": "3.4.0",
 		"phpunit/phpunit": "^9",
 		"phpstan/phpstan-php-parser": "^1.0",
 		"phpstan/phpstan-strict-rules": "^1.1",
-		"phpstan/phpstan-phpunit": "^1.0",
-		"symplify/phpstan-extensions": "^10.0"
```
staabm, line 16: unused dependency

PR #49, opened 2022-01-08
remove `false` from return types, when error mode is exception

staabm:
see also https://github.com/php/doc-en/pull/1313

- when analyzing a php8+ codebase, [`PDO::ERRMODE_EXCEPTION` error handling](https://www.php.net/manual/en/pdo.error-handling.php) is assumed.
- when analyzing a php8.1+ codebase, [`mysqli_report(\MYSQLI_REPORT_ERROR | \MYSQLI_REPORT_STRICT);` error handling](https://www.php.net/mysqli_report) is assumed.


staabm:
todo: different return-type expectations for mysql on php 8.0 vs 8.1

PR #52, opened 2022-01-09
support null and string-values in prepared statements

staabm:
see https://stackoverflow.com/questions/48643967/php-pdo-prepared-statement-bind-null-value

### DIFF
--- a/tests/data/pdo-stmt-execute.php
+++ b/tests/data/pdo-stmt-execute.php
@@ -19,6 +19,11 @@ class Foo
         $stmt->execute(['adaid' => 1]); // prefixed ":" is optional
         assertType('PDOStatement<array{email: string, 0: string, adaid: int<0, 4294967295>, 1: int<0, 4294967295>}>', $stmt);
 
+        $stmt = $pdo->prepare('SELECT email, adaid FROM ada WHERE email = :email');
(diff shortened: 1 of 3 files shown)


tests/data/pdo-stmt-execute.php
```diff
@@ -19,10 +19,25 @@ public function execute(PDO $pdo)
         $stmt->execute(['adaid' => 1]); // prefixed ":" is optional
         assertType('PDOStatement<array{email: string, 0: string, adaid: int<0, 4294967295>, 1: int<0, 4294967295>}>', $stmt);
 
+        $stmt = $pdo->prepare('SELECT email, adaid FROM ada WHERE email = :email');
+        assertType('PDOStatement', $stmt);
+        $stmt->execute([':email' => 'email@example.org']);
+        assertType('PDOStatement<array{email: string, 0: string, adaid: int<0, 4294967295>, 1: int<0, 4294967295>}>', $stmt);
+
+        $stmt = $pdo->prepare('SELECT email, adaid FROM ada WHERE email <=> :email');
```
staabm, line 27: mysql syntax für NULL values in prepared statements, see https://dev.mysql.com/doc/refman/5.7/en/comparison-operators.html#operator_equal-to

PR #65, opened 2022-01-10
wip

staabm:


staabm:
resumè: the experiement shows, we need a full query parser to find out, which named-parameter is assigned to which db-column, so we can compare column-types against param-value-types

PR #66, opened 2022-01-10
Support `\Doctrine\DBAL\Connection->query()` syntax error checking

staabm:
Based on https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#data-retrieval-and-manipulation

### DIFF
--- a/config/dba.neon
+++ b/config/dba.neon
@@ -15,6 +15,7 @@ services:
 				- 'PDO::query#0'
 				- 'PDO::prepare#0'
 				- 'mysqli::query#0'
+				- 'Doctrine\DBAL\Connection::query#0'
 
 	-
 		class: staabm\PHPStanDba\Rules\SyntaxErrorInQueryFunctionRule
--- a/.phpstan-dba.cache
+++ b/.phpstan-dba.cache
@@ -213,7 +213,6 @@
       'result' => 
       array (
         1 => NULL,
-        2 => NULL,
         3 => NULL,
       ),
     ),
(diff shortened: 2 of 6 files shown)


config/dba.neon
```diff
@@ -15,6 +15,7 @@ services:
 				- 'PDO::query#0'
 				- 'PDO::prepare#0'
 				- 'mysqli::query#0'
+				- 'Doctrine\DBAL\Connection::query#0'
```
Seldaek, line 18: This is actually deprecated. `executeQuery($sql, $params)`, `executeCacheQuery($sql, $params)`, `executeStatement($sql, $params)` and `prepare($sql)` are the current methods available on [Doctrine\DBAL\Connection](https://github.com/doctrine/dbal/blob/3.2.x/src/Connection.php), but `executeUpdate`, `exec` and `query` are still supported and widely used so probably I'd keep them in as well.

There's also `fetchAllNumeric`, `fetchAllNumeric`, `fetchAllAssociative`, `fetchAllKeyValue`, `fetchAllAssociativeIndexed`, `fetchFirstColumn`, `iterateNumeric`, `iterateAssociative`, `iterateKeyValue`, `iterateAssociativeIndexed`, `iterateColumn` which are all shorthands for calling these on the returned PDOStatement/Result.

I'm not sure if it's as easy as listing all of them here? I guess it'd do it for the query, but what about checking $params? This is currently only working for direct PDO statements via PdoStatementExecuteMethodRule above is that correct? I guess this is more involved and would require a new rule?

staabm: for prepared-statement style methods which take a $params array there is a `https://github.com/staabm/phpstan-dba/blob/main/src/Rules/SyntaxErrorInPreparedStatementMethodRule.php` rule which can be configured.

this rule expects that the query is passed as a 1st arg and params as a 2nd. but this seems to fit for `executeQuery($sql, $params)`, `executeCacheQuery($sql, $params)`, `executeStatement($sql, $params)`

> There's also `fetchAllNumeric`, `fetchAllNumeric`, `fetchAllAssociative`, `fetchAllKeyValue`, `fetchAllAssociativeIndexed`, `fetchFirstColumn`, `iterateNumeric`, `iterateAssociative`, `iterateKeyValue`, `iterateAssociativeIndexed`, `iterateColumn` which are all shorthands for calling these on the returned PDOStatement/Result.

I think for these methods we need custom return type extensions similar to https://github.com/staabm/phpstan-dba/blob/main/src/Extensions/PdoStatementFetchDynamicReturnTypeExtension.php

Seldaek: > I think for these methods we need custom return type extensions similar to https://github.com/staabm/phpstan-dba/blob/main/src/Extensions/PdoStatementFetchDynamicReturnTypeExtension.php

Yes absolutely for the return types, I was more concerned about getting the SQL/params validated everywhere. I think these could also be added to SyntaxErrorInPreparedStatementMethodRule.php then? All of them have the same basic arg signature.

staabm: I will have a look into the extensions-side

staabm: > Yes absolutely for the return types, I was more concerned about getting the SQL/params validated everywhere. I think these could also be added to SyntaxErrorInPreparedStatementMethodRule.php then? All of them have the same basic arg signature.

exactly. just add the rule to the dba.neon config and list them.

the rule is not yet listed in dba.neon, as there is not yet a builtin method for this rule.
I am using it in a closed project already though for a different api. so it should work.

staabm: see https://github.com/staabm/phpstan-dba/pull/116 for syntax error detection

PR #79, opened 2022-01-11
fix `RecordingQueryReflector` for prepared statements

staabm:


tests/data/syntax-error-in-query.php
```diff
@@ -62,14 +62,6 @@ public function validQuery(PDO $pdo)
         $pdo->query('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada', PDO::FETCH_ASSOC);
     }
 
-    public function queryWithPlaceholder(PDO $pdo)
-    {
-        // atm we just make sure, this is not detected as a syntax error.
-        // a proper return type check should be added in the future.
-        $pdo->query('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada WHERE adaid=?', PDO::FETCH_ASSOC);
-        $pdo->query('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada WHERE adaid=:adaid', PDO::FETCH_ASSOC);
```
staabm, line 70: we have proper coverage for this cases already in `pdo-stmt-execute.php`

PR #87, opened 2022-01-11
added cache to reduce number of execute queries

staabm:


staabm:
superceeded by https://github.com/staabm/phpstan-dba/pull/103

Issue #69, opened 2022-01-10
Prepared statement: Question mark parameters


craigfrancis:
Currently `phpstan-dba` supports named parameters (thank you), but it does not support question mark parameters:

    $statement = $pdo->prepare('SELECT * FROM user WHERE id = :id');
    $statement->execute(['id' => 1]);
    
    $statement = $pdo->prepare('SELECT * FROM user WHERE id = ?');
    $statement->execute([1]);

First one is fine, second one "Value 0 is given to execute(), but the query does not containt this placeholder."

I can see you have a `$parameters` array in `PdoStatementExecuteErrorMethodRule::checkParameterValues()`, is it worth using something like the following to determine if it's using keyed arrays:

    (count(array_filter(array_keys($array), 'is_string')) > 0)

https://stackoverflow.com/a/4254008/6632

---

Also, typo with "containt" (says he with a keyboard who happens to have a nearly broken "t" key, so can you send them my way).

staabm:
> but it does not support question mark parameters:

hmm question mark placeholders are already supported, see [unit test](https://github.com/staabm/phpstan-dba/blob/81338e46db23b583539d77456fc3b0809a620505/tests/data/pdo-stmt-execute.php#L32-L35)

could you provide a failling test-case?

craigfrancis:
I'll try to work out what's happening later... I've just run an update, and tested this simple script:

```
<?php

$pdo = new PDO('mysql:dbname=test;host=localhost', 'test', 'test', [PDO::ATTR_EMULATE_PREPARES => false]);

$statement = $pdo->prepare('SELECT * FROM user WHERE id = ?');
$statement->execute([1]);

?>
```

And it errors with:

```
 ------ ---------------------------------------------------------------------- 
  Line   index.php                                                             
 ------ ---------------------------------------------------------------------- 
  5      Query error: You have an error in your SQL syntax; check the manual   
         that corresponds to your MariaDB server version for the right syntax  
         to use near '? LIMIT 0' at line 1 (1064).                             
  6      Value 0 is given to execute(), but the query does not contain this    
         placeholder.                                                          
 ------ ---------------------------------------------------------------------- 
```

I've been thrown by the new `LIMIT 0` error.

[...]

PR #105, opened 2022-01-16
Do not run PDO namedPlaceholders checks with question mark parameters

craigfrancis:
Adding a simple `if (\count($namedPlaceholders) > 0) {` around the **named** placeholders checks.

This stops the following error when checking `$stmt->execute()` on a query using question mark parameters.

    Value 0 is given to execute(), but the query does not contain this placeholder.

This is for [Issue 69](https://github.com/staabm/phpstan-dba/issues/69)

### DIFF
--- a/src/Rules/PdoStatementExecuteMethodRule.php
+++ b/src/Rules/PdoStatementExecuteMethodRule.php
@@ -122,15 +122,17 @@ final class PdoStatementExecuteMethodRule implements Rule
 
         $errors = [];
         $namedPlaceholders = $this->extractNamedPlaceholders($queryString);
-        foreach ($namedPlaceholders as $namedPlaceholder) {
-            if (!\array_key_exists($namedPlaceholder, $parameters)) {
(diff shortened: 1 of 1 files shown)


staabm:
Hmm the only way that `extractNamedPlaceholders` might not return an array is

is this path: https://github.com/staabm/phpstan-dba/blob/7c3d0701af2e24058e8249716e4ba69ffd9068ee/src/Rules/PdoStatementExecuteMethodRule.php#L174

seems there is something wrong with this if-case. Could you look into the root instead of adding this workaround?

alternatively a failling unit test would be great

craigfrancis:
It's returning an array, it's just that the array is empty, on [line 170](https://github.com/staabm/phpstan-dba/blob/7c3d0701af2e24058e8249716e4ba69ffd9068ee/src/Rules/PdoStatementExecuteMethodRule.php#L170).

staabm (approved):
Ah I get it now. Thanks - look good

PR #110, opened 2022-01-17
Add support for JSON columns and nullability for string column types

Seldaek:
While I'm here, for DATE/DATETIME I guess string is also the only reasonable output.. unless PHPStan has a concept of date-string (which would be kinda neat;)


Seldaek:
I guess regarding nullability there is one more complicating aspect.. if you do a LEFT JOIN then any column in the joined table becomes nullable. Is this already taken care of? :s

staabm:
> if you do a LEFT JOIN then any column in the joined table becomes nullable

just verified that the left-join nullability is already supported because of the result-set type inference.

added a test for that in https://github.com/staabm/phpstan-dba/pull/126

PR #111, opened 2022-01-17
Support complete placeholder name matching

### LINKED ISSUE #109: Prepared statements with camelCased named placeholders
Not sure if related to another of the open issues tbh, but I tried this as it's IMO very promising and had to go around Doctrine DBAL for now to test it out but I got a rather unexpected result:

```
        $sql = '
            SELECT v.normalizedVersion, d.data
            FROM package_version v
            INNER JOIN download d ON d.id=v.id AND d.type = :versionType
            WHERE v.package_id = :package AND v.development = 0 AND v.normalizedVersion LIKE :majorVersion
        ';

        /** @var \Doctrine\DBAL\Driver\PDO\Connection $pdo */
        $pdo = $this->getEntityManager()->getConnection()->getWrappedConnection();
        $stmt = $pdo->prepare($sql);
        $stmt->execute(['package' => $package->getId(), 'versionType' => Download::TYPE_VERSION, 'majorVersion' => $majorVersion . '.%']);

        $result = $stmt->fetchAll(PDO::FETCH_ASSOC);
```

Line 37 is the `execute` call.
```
  37     Query expects placeholder :major, but it is missing from values given to execute().
  37     Query expects placeholder :version, but it is missing from values given to execute().
  37     Value :majorVersion is given to execute(), but the query does not contain this placeholder.
  37     Value :versionType is given to execute(), but the query does not contain this placeholder.
```

I have no idea why the placeholders `:majorVersion`/`:versionType` became `:major`/`:version`. It looks like it stopped parsing as soon as it encountered a capital letter, so it smells to me of a regex that's not case insensitive or something.. But I couldn't find anything in the code.

Seldaek:
Fixes #109

As per https://github.com/php/php-src/blob/01b3fc03c30c6cb85038250bb5640be3a09c6a32/ext/pdo/pdo_sql_parser.re#L48 I believe this is now correct/complete :)

### DIFF
--- a/.phpstan-dba.cache
+++ b/.phpstan-dba.cache
@@ -212,8 +212,8 @@
       )),
       'result' => 
       array (
-        2 => NULL,
         1 => NULL,
+        2 => NULL,
         3 => NULL,
       ),
     ),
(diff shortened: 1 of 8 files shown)


staabm:
perfect, thx

PR #112, opened 2022-01-17
extract REGEX_PLACEHOLDER

staabm:
closes https://github.com/staabm/phpstan-dba/issues/109


staabm:
will fix in https://github.com/staabm/phpstan-dba/pull/111

Issue #94, opened 2022-01-12
Question about Doctrine/DBAL


mitelg:
Hey there :wave: 

I found your project on twitter the other day and after looking into it, it looks very interesting. :+1: good work so far

we are using the Doctrine/DBAL connection in our projects and not directly the PHP classes and functions. Is this somehow also supported? Or are your interfered types again replaced by the annotations and types of DBAL itself? 

keep up the good work :muscle:

staabm:
Hi.

thx for the feedback.

whats already possible today is re-using of `SyntaxErrorInPreparedStatementMethodRule` to get syntax error checking capabilities for queries which are known at analysis time - [see readme](https://github.com/staabm/phpstan-dba#use-syntaxerrorinpreparedstatementmethodrule-for-your-custom-classes)

when time allows I am planning to add this by default with https://github.com/staabm/phpstan-dba/pull/66 (I don't use doctrine/dbal myself, so priorties are on other stuff)

----

what could be done on top of that is making phpstan aware of methods/functions/types involved while building query strings.
for example, if we would have phpstan extensions for the query builder api, queries like
```
$queryBuilder
    ->select('id', 'name')
    ->from('users')
    ->where('email = ?')
    ->setParameter(0, $userInputEmail)
;
```

could also be analyzable.

----

> Or are your interfered types again replaced by the annotations and types of DBAL itself?

as phpstan-dba already works on the schema available in the database, I don't see value atm in reading the doctrine annotations. I think, if we would have the above mentioned steps implemented this might change though.

mitelg:
that looks promising :+1: 

just FYI: https://github.com/phpstan/phpstan-doctrine/blob/master/src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
in the PHPStan doctrine plugin, there is already a small evaluation of the Querybuilder methods, but it is only about what is `execute` is giving back. 

I will try your extension, once I will find some time :grin:

staabm:
just implemented this idea with zend framework 1 query builder:
https://twitter.com/markusstaab/status/1482999683911987201

mitelg:
hey @staabm 

I tried your current master on our project https://github.com/shopware/shopware

there were some issues found and also some errors during the execution :grin:

I guess these things need to be fixed on your side:

[...]

PR #115, opened 2022-01-17
Added initial Doctrine Dbal type inference

staabm:

adds support for
- `Connection->query()`
- `Result->fetchNumeric()`
- `Result->fetchAssociative()`
- `Result->fetchAllNumeric()`
- `Result->fetchAllAssociative()`
- `Result->columnCount()`

### DIFF
--- a/src/Extensions/DoctrineResultDynamicReturnTypeExtension.php
+++ b/src/Extensions/DoctrineResultDynamicReturnTypeExtension.php
@@ -17,7 +17,7 @@ use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\Generic\GenericObjectType;
-use PHPStan\Type\IntegerType;
+use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\Type;
 use staabm\PHPStanDba\QueryReflection\QueryReflector;
 
@@ -91,7 +91,7 @@ final class DoctrineResultDynamicReturnTypeExtension implements DynamicMethodRet
             }
 
             if (\in_array(strtolower($methodReflection->getName()), ['fetchallnumeric', 'fetchallassociative'], true)) {
-                return new ArrayType(new IntegerType(), $builder->getArray());
+                return new ArrayType(IntegerRangeType::fromInterval(0, null), $builder->getArray());
             }
 
             return $builder->getArray();
--- a/src/Extensions/DoctrineResultDynamicReturnTypeExtension.php
+++ b/src/Extensions/DoctrineResultDynamicReturnTypeExtension.php
@@ -9,12 +9,14 @@ use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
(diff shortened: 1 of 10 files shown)


tests/data/doctrine-dbal.php
```diff
@@ -0,0 +1,21 @@
...
+    public function foo(Connection $conn)
+    {
+        $result = $conn->query('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada');
+        assertType('Doctrine\DBAL\Result<array{email: string, 0: string, adaid: int<0, 4294967295>, 1: int<0, 4294967295>, gesperrt: int<-128, 127>, 2: int<-128, 127>, freigabe1u1: int<-128, 127>, 3: int<-128, 127>}>', $result);
+
+        $numericFetch = $result->fetchNumeric();
+        assertType('array{string, int<0, 4294967295>, int<-128, 127>, int<-128, 127>}', $numericFetch);
+
+        $numericFetch = $result->fetchAssociative();
+        assertType('array{email: string, adaid: int<0, 4294967295>, gesperrt: int<-128, 127>, freigabe1u1: int<-128, 127>}', $numericFetch);
```
staabm, line 19: @Seldaek could you verify this api and type expectations are correct?

src/Extensions/DoctrineResultDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,102 @@
...
+        if (1 !== \count($genericTypes)) {
+            return $defaultReturn;
+        }
+
+        $resultRowType = $genericTypes[0];
+
+        switch (strtolower($methodReflection->getName())) {
+            case 'columncount':
+                if ($resultRowType instanceof ConstantArrayType) {
+                    $columnCount = \count($resultRowType->getKeyTypes()) / 2;
```
Seldaek, line 57: I'm not sure if you can simply divide here, it may depend on the FETCH_TYPE used when executing the statement? As in if all keys are numeric or all keys are associative, then you shouldn't divide by 2? Unless I misunderstand what is going on here.. :)

Seldaek: Ah I guess in a static analysis context this is not true as you always have all columns listed. Never mind.

tests/data/doctrine-dbal.php
```diff
@@ -0,0 +1,30 @@
...
+        assertType('array{string, int<0, 4294967295>, int<-128, 127>, int<-128, 127>}', $fetch);
+
+        $fetch = $result->fetchAssociative();
+        assertType('array{email: string, adaid: int<0, 4294967295>, gesperrt: int<-128, 127>, freigabe1u1: int<-128, 127>}', $fetch);
+
+        $fetch = $result->fetchAllNumeric();
+        assertType('array<int, array{string, int<0, 4294967295>, int<-128, 127>, int<-128, 127>}>', $fetch);
+
+        $fetch = $result->fetchAllAssociative();
+        assertType('array<int, array{email: string, adaid: int<0, 4294967295>, gesperrt: int<-128, 127>, freigabe1u1: int<-128, 127>}>', $fetch);
```
Seldaek, line 25: fetchAll* should perhaps return `list<array{}>` ? I imagine this is true for all drivers but who knows.

staabm: phpstan does not yet support the list type natively. atm you can do this in phpdoc, but internally its still a array.

I thought about using a `positive-int` key.. not sure this holds?

Seldaek: Ah interesting :) Yes I'd think positive-int is true, if list<> is true.. I'm sure it's 99% true and there's someone with a weird implementation somewhere that breaks it but probably this won't matter?

staabm: yeah, I think we should go as strict as possible.

staabm:
@Seldaek I have no application to actually test this extension. could you give it a go for a small smoketest?

if it works for you, we can ship the next release later today with initial doctrine support :-)

Seldaek:
Yeah happy to check later but probably easier if you merge it all first.

tests/data/doctrine-dbal.php
stof: be careful. `positive-int` means `>0`, which is wrong for lists

staabm: Thx. We used int<0, max>

PR #116, opened 2022-01-17
Detect syntax errors in `Doctrine\DBAL\Connection::execute*()` methods

staabm:
This adds support for `executeQuery($sql, $params)`, `executeCacheQuery($sql, $params)`, `executeStatement($sql, $params)` 

`prepare($sql)` is not yet supported and can be implemented later

### DIFF
--- a/config/dba.neon
+++ b/config/dba.neon
@@ -1,26 +1,35 @@
 includes:
-	- stubFiles.neon
-	- extensions.neon
+    - stubFiles.neon
+    - extensions.neon
 
 services:
-	-
-		class: staabm\PHPStanDba\Rules\PdoStatementExecuteMethodRule
-		tags: [phpstan.rules.rule]
+    -
+        class: staabm\PHPStanDba\Rules\SyntaxErrorInPreparedStatementMethodRule
+        tags: [phpstan.rules.rule]
+        arguments:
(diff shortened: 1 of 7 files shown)


config/dba.neon
```diff
@@ -1,26 +1,35 @@
...
-				- 'mysqli_query#1'
+    -
+        class: staabm\PHPStanDba\Rules\SyntaxErrorInQueryMethodRule
+        tags: [phpstan.rules.rule]
+        arguments:
+            classMethods:
+                - 'PDO::query#0'
+                - 'PDO::prepare#0'
+                - 'mysqli::query#0'
+                - 'Doctrine\DBAL\Connection::query#0'
```
Seldaek, line 27: IMO you could remove `query` and only support the latest APIs. Or you keep it but then I'd also add `exec` here and `executeUpdate` above in `SyntaxErrorInPreparedStatementMethodRule` for completeness.

Seldaek:
Oh, too slow ;) Thanks for all the good work here!

config/dba.neon
staabm: correct.. its just a few lines of config.. lets support them: https://github.com/staabm/phpstan-dba/pull/117

PR #151, opened 2022-01-19
prevent wrong type inference on doctrine dbal 2.x

staabm:
refs https://github.com/staabm/phpstan-dba/issues/94#issuecomment-1016534736

should prevent 

> * `\Doctrine\DBAL\Connection::query` returns `\Doctrine\DBAL\Driver\Statement` and not `Doctrine\DBAL\Result`. I get errors like this: `Call to an undefined method Doctrine\DBAL\Result<array<int|string, int>>::fetchColumn().`
> we use version 2.13

uses https://getcomposer.org/doc/07-runtime.md#installed-versions


staabm:
@mitelg could you check whether this PR fixes the errors you see regarding `Call to an undefined method Doctrine\DBAL\Result` ?

src/Extensions/DoctrineConnectionDynamicReturnTypeExtension.php
```diff
@@ -4,6 +4,8 @@
 
 namespace staabm\PHPStanDba\Extensions;
 
+use Composer\InstalledVersions;
+use Composer\Semver\VersionParser;
```
Seldaek, line 8: You need to require composer/semver if you want to use `satisfies`.

composer.json
```diff
@@ -5,6 +5,7 @@
 	"require": {
 		"php": "^8.0",
 		"ext-mysqli": "*",
+		"composer-runtime-api": "^2.2",
```
staabm, line 8: 2.0 should be enough

src/Extensions/DoctrineConnectionDynamicReturnTypeExtension.php
staabm: Thx for the hint. Interesting that the build is green even though the dependency is not declared

Seldaek: Probably something else you depend on requires it, so it works but not super failsafe :)

mitelg:
code wise this looks good, but didn't had the chance yet, to check it

PR #153, opened 2022-01-19
don't report placeholder errors on non-SELECT statements for now

staabm:
with proper sql parsing, we should better detect the placeholders and therefore could validate this query

refs https://github.com/staabm/phpstan-dba/issues/94#issuecomment-1016534736

### DIFF
--- a/tests/data/syntax-error-in-prepared-statement.php
+++ b/tests/data/syntax-error-in-prepared-statement.php
@@ -174,7 +174,8 @@ class Foo
         $connection->preparedQuery($query, ['adaids' => [1, 2, 3]]);
     }
 
-    public function noErrorInBug94(Connection $connection) {
+    public function noErrorInBug94(Connection $connection)
+    {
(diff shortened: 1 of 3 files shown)


tests/data/syntax-error-in-prepared-statement.php
```diff
@@ -173,4 +173,30 @@ public function arrayParam(Connection $connection)
         $query = 'SELECT email FROM ada WHERE adaid IN (:adaids)';
         $connection->preparedQuery($query, ['adaids' => [1, 2, 3]]);
     }
+
+    public function noErrorInBug94(Connection $connection)
+    {
+        // XXX with proper sql parsing, we should better detect the placeholders and therefore could validate this query
+        $sql = "
+            INSERT IGNORE INTO `s_articles_supplier` (`id`, `name`, `img`, `link`, `changed`) VALUES (:supplierId, 'TestSupplier', '', '', '2019-12-09 10:42:10');
```
staabm, line 181: the current placeholder parsing detects `:42` of the datetime-col as a placeholder

Issue #174, opened 2022-01-21
unexpected error in query analysis


staabm:
refs https://github.com/composer/packagist/pull/1232

```php
    /**
     * @param string   $name Package name to find the dependents of
     * @param int|null $type One of Dependent::TYPE_*
     * @return int<0, max>
     */
    public function getDependentCount(string $name, ?int $type = null): int
    {
        $sql = 'SELECT COUNT(*) count FROM dependent WHERE packageName = :name';
        $args = ['name' => $name];
        if (null !== $type) {
            $sql .= ' AND type = :type';
            $args['type'] = $type;
        }

        return (int) $this->getEntityManager()->getConnection()->fetchOne($sql, $args);
    }
```

leads to 

> Query expects 1 placeholder, but 2 values are given. 

no error expected

PR #181, opened 2022-01-22
support optional parameter values

staabm:
depends on https://github.com/phpstan/phpstan/issues/6439

closes https://github.com/staabm/phpstan-dba/issues/174


staabm:
The linked phpstan issue might add support for the unit test but would still lead a similar problem we see atm, when we get combinatorial complex enough that phpstan in the end will only return a intersection type instead of the literal query string

PR #184, opened 2022-01-23
added support for unions mixing non-simulatable and simulatable types

staabm:


tests/SyntaxErrorInQueryMethodRuleTest.php
```diff
@@ -50,7 +50,7 @@ public function testSyntaxErrorInQueryRule(): void
                 41,
             ],
             [
-                "Query error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL/MariaDB server version for the right syntax to use near 'FROM ada  WHERE email='my_other_table' LIMIT 0' at line 1 (1064).",
+                "Query error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL/MariaDB server version for the right syntax to use near 'FROM ada LIMIT 0' at line 1 (1064).",
```
staabm, line 53: before this fix we used the last supported union-element-type to build the error message.
now we return early, use the first union-element-type and therefore have a different message.

PR #234, opened 2022-02-03
SyntaxErrorInQueryMethodRule: allow abstract classes in "classMethods"

voku:
Sorry, I haven't much time currently, so I changed this via github and without any tests. :|

### DIFF
--- a/src/Rules/SyntaxErrorInQueryMethodRule.php
+++ b/src/Rules/SyntaxErrorInQueryMethodRule.php
@@ -27,17 +27,11 @@ final class SyntaxErrorInQueryMethodRule implements Rule
     private $classMethods;
 
     /**
-     * @var ReflectionProvider
-     */
-    private $reflectionProvider;
-
-    /**
      * @param list<string> $classMethods
      */
-    public function __construct(array $classMethods, ReflectionProvider $reflectionProvider)
+    public function __construct(array $classMethods)
     {
         $this->classMethods = $classMethods;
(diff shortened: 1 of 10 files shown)


staabm:
it seems you are trying to subclass `SyntaxErrorInQueryMethodRule`. could you describe your use-case?

`SyntaxErrorInQueryMethodRule` is declared `final` atm, so this won't work atm I guess.

src/Rules/SyntaxErrorInQueryMethodRule.php
```diff
@@ -54,7 +54,7 @@ public function processNode(Node $node, Scope $scope): array
         foreach ($this->classMethods as $classMethod) {
             sscanf($classMethod, '%[^::]::%[^#]#%s', $className, $methodName, $queryArgPosition);
 
-            if ($methodName === $methodReflection->getName() && $className === $methodReflection->getDeclaringClass()->getName()) {
+            if ($methodName === $methodReflection->getName() && is_a($className, $methodReflection->getDeclaringClass()->getName(), true)) {
```
staabm, line 57: I finally got the idea what you are trying to achieve here.

With `is_a` this check would require runtime reflection.
Instead it should use static reflection
https://github.com/phpstan/phpstan-src/blob/d5e6df75d42ad942f0cd1d592c0faf6a30c85c5d/src/Reflection/ClassReflection.php#L599

staabm:
today we found the very same use-case within one of our apps - @voku it seems your apps have the same design flaws since ours :)

PR #239, opened 2022-02-04
support queries containing `implode()` of constant values

staabm:
will be fixed with https://github.com/phpstan/phpstan-src/pull/991


staabm:
needs re-record.. otherwise fixed

Issue #225, opened 2022-02-02
Strange DBAL discrepancy between local env and CI


Seldaek:
With this code here https://github.com/composer/packagist/blob/8a3d28e42f9790dea401e5f9afc7c2c9780b58d8/src/Entity/PackageRepository.php#L255-L270

Running locally I get:

> Method App\Entity\PackageRepository::getStalePackagesForDumping() should return array<int, string> but returns array<int<0, max>, int<0, 4294967295>>.

Which makes sense given #222

But on CI e.g. https://github.com/composer/packagist/runs/5037284175?check_suite_focus=true I get this:

> Method App\Entity\PackageRepository::getStalePackagesForDumping() should return array<int, string> but returns array<int, mixed>.

Any clue what could cause this? All dependency versions are the same as it installs from lock file.

staabm:
having a mixed only for a single column-type would mean you are running into this case:
https://github.com/staabm/phpstan-dba/blob/e2939f28cb3e2b8cc465c15e78b965eade5b11c3/src/QueryReflection/MysqliQueryReflector.php#L293

are you sure you are using the same db schema locally and in CI?
maybe its a difference in mysql versions or maria db vs. mysql?

Seldaek:
The package table has a simple int(11) primary id, nothing weird, and it definitely hasn't changed in 10 years so I believe my local schema matches CI :) 

https://github.com/composer/packagist/blob/8a3d28e42f9790dea401e5f9afc7c2c9780b58d8/src/Entity/Package.php#L62-L67

I'm on mysql8.0.13 and GH is on 8.0.26 but I doubt that'd make a difference here.. I can try to upgrade to be sure, wouldn't hurt anyway ;)

staabm:
If time allows I can debug the CI job and the used phpstan-dba version within the CI pipeline.

Seldaek:
On 8.0.28 locally now, still can't repro :/

Seldaek:
Very confused, just tried again to run phpstan and suddenly it changed to the same output as CI?! 

I even tried clearing PHPStan & phpstan-dba caches earlier, it didn't help.. proper WTF here.. but at least it works now so I guess closing 🤷‍♂️

staabm:
hm ok, so you are now able to debug locally why its returning `mixed`?

running phpstan with `--debug` disables all caches.. thats the first thing I try when caching might be a problem.

staabm:
[...]

PR #243, opened 2022-02-04
debug-mode: report error when query is not simulatable

staabm:
to ease debugging of errors like https://github.com/staabm/phpstan-dba/issues/225#issuecomment-1029970147

### DIFF
--- a/src/QueryReflection/MysqliQueryReflector.php
+++ b/src/QueryReflection/MysqliQueryReflector.php
@@ -21,8 +21,8 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
 use staabm\PHPStanDba\Error;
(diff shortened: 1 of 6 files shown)


src/QueryReflection/MysqliQueryReflector.php
```diff
@@ -119,6 +119,10 @@ public function getResultType(string $queryString, int $fetchType): ?Type
     {
         $result = $this->simulateQuery($queryString);
         if (!\is_array($result)) {
+            if (QueryReflection::getRuntimeConfiguration()->isDebugEnabled() && $result instanceof mysqli_sql_exception) {
+                throw new UnresolvableQueryException(sprintf("Cannot simulate query\n %s \nbecause of a sql error: %s", $queryString, $result->getMessage()));
```
staabm, line 123: todo: separate exception type

staabm:
Todo: make this report a phpstan error instead of a uncaught exception

Issue #253, opened 2022-02-06
PDO: Support setting default fetch mode


xPaw:
It is possible to configure the default fetch mode in PDO:

```php
new PDO(..., [
  PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_ASSOC,
]);
```

It would be similar to setting errorMode on RuntimeConfiguration.

staabm:
Could you give a full example how could beeing analyzed would look like when using this feature?

xPaw:
The new issue you created would be the same example.

```php
<?php
$dbh = new PDO(..., [PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_NUM]);

$stmt = $dbh->query('SELECT name, colour, calories FROM fruit');
foreach ($stmt as $row) {
    print $row[0] . "\t" . $row[1] . "\t" . $row[2] . "\n";
}
```

staabm:
Thx. A new runtime configuration option would be a perfect fit.

are you willing to contribute?

xPaw:
I'll take a look.

PR #256, opened 2022-02-06
Add `RuntimeConfiguration->defaultFetchMode()` option

xPaw:
A few notes that may need to be looked at in the future:

- Query/statement code essentially have duplicate fetch mode lookup code.
- Not all fetch modes are supported.
- Adding new tests for each runtime config requires duplicating a lot of the boilerplate (which is why I didn't bother copying the folder for every fetch type). It probably would be beneficial to allow tests changing runtime options without having a new bootstrap.php file.
- Tests default to hostname that doesn't resolve.

closes https://github.com/staabm/phpstan-dba/issues/253


tests/schema.sql
```diff
@@ -1,4 +1,4 @@
-CREATE DATABASE IF NOT EXISTS `phpstan-dba`;
+CREATE DATABASE IF NOT EXISTS `phpstan_dba`;
```
xPaw, line 1: How did this work?

staabm: it seems we created a database with this schema file, but used a different one :).

but you already figured that one out as far as I can see. good catch

staabm:
thanks this PR looks good.

I did a small refactoring on the `main` branch, therefore needs a rebase. I can take care of it, if you don't have time doing so.

> Query/statement code essentially have duplicate fetch mode lookup code.

yeah we could DRY a bit.. but atm I still try to find a good api, therefore I left it as is.

> Adding new tests for each runtime config requires duplicating a lot of the boilerplate (which is why I didn't bother copying the folder for every fetch type). It probably would be beneficial to allow tests changing runtime options without having a new bootstrap.php file.

since the runtime-config is kind of a global config I am fine how it works atm.
hopefully we won't need too much options :)

src/QueryReflection/RuntimeConfiguration.php
```diff
@@ -53,6 +58,18 @@ public function errorMode(string $mode): self
         return $this;
     }
 
+    /**
+     * Defines default fetch mode.
+     *
+     * @param QueryReflector::FETCH_TYPE* $mode
+     */
+    public function fetchMode(int $mode): self
```
staabm, line 66: I will rename that one to `defaultFetchMode`, as it will only be the fallback for cases no fetch-mode is defined otherwise

xPaw:
Rebased.

xPaw:
It seems that type for `$result = $stmt->fetch(PDO::FETCH_NUM);` gets lost and its just `array{}` unsure if I broke it in this PR (wouldn't other tests fail though?)

staabm:
the point is, that the `PDOStatement` itself needs to hold all possible generic types (atm its `BOTH`).

from such a "generic-PDOStatement" we can reduce the typing to a more specialized form using `PdoStatementReflection->getStatementResultType()`.

to sum up: in the past we always used `BOTH` at the first step and then reduced the type from their. I think we still need to do it like that (even if the defaultFetchMode is no longer `BOTH`). does it make sense?

xPaw:
I see what you mean, but I'm not entirely sure how to accomplish that, e.g. for this case:

```
        $stmt = $pdo->prepare('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada');
        assertType('PDOStatement<array{email: string, adaid: int<0, 4294967295>, gesperrt: int<-128, 127>, freigabe1u1: int<-128, 127>}>', $stmt);
```

staabm:
the alternative would be that we would try to use `findPrepareQueryStringExpression` in case the generic of the current `PDOStatement` does not contain the necessary information, because of a different fetch-type used initially

xPaw:
I think `PdoPrepareDynamicReturnTypeExtension` using the default type is correct, it probably needs to work very closely with the code you added in `setFetchMode` PR.

Unfortunately I can't come up with a working solution right now.

tests/fetchMode/data/modes.php
```diff
@@ -0,0 +1,53 @@
...
+    public function assocModeFetchOverriden(PDO $pdo)
+    {
+        // default fetch-type is globally changed to assoc for this test-suite
+
+        $stmt = $pdo->prepare('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada');
+        assertType('PDOStatement<array{email: string, adaid: int<0, 4294967295>, gesperrt: int<-128, 127>, freigabe1u1: int<-128, 127>}>', $stmt);
+        $stmt->execute();
+        assertType('PDOStatement<array{email: string, adaid: int<0, 4294967295>, gesperrt: int<-128, 127>, freigabe1u1: int<-128, 127>}>', $stmt);
+        $result = $stmt->fetch(PDO::FETCH_NUM);
+        assertType('array{email: string, adaid: int<0, 4294967295>, gesperrt: int<-128, 127>, freigabe1u1: int<-128, 127>}', $result);
```
xPaw, line 41: fyi this type is not correct at the moment, and it probably also needs a FETCH_BOTH test here.

tests/fetchMode/data/modes.php
```diff
@@ -0,0 +1,53 @@
...
+use PDO;
+use function PHPStan\Testing\assertType;
+
+class Foo
+{
+    public function assocModeQuery(PDO $pdo)
+    {
+        // default fetch-type is globally changed to assoc for this test-suite
+
+        $stmt = $pdo->query('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada');
```
staabm, line 14: there is another test-case when passing the `fetch-type` as a 2nd arg to `query()`

src/QueryReflection/RuntimeConfiguration.php
```diff
@@ -53,6 +57,19 @@ public function errorMode(string $mode): self
         return $this;
     }
 
+    /**
+     * Defines the PDO default fetch mode.
+     * This might be necessary in case you are using `\PDO::ATTR_DEFAULT_FETCH_MODE`.
+     *
+     * @param QueryReflector::FETCH_TYPE_BOTH|QueryReflector::FETCH_TYPE_ASSOC $mode
```
staabm, line 64: to simplify we only allow a default-fetch-mode of BOTH or ASSOC for now.

@xPaw as I understood it, this is sufficient for your use case?

xPaw: Looking at my code base, my actual use case is setting ASSOC by default on the PDO object, and then sometimes using PDO::FETCH_KEY_PAIR, PDO::FETCH_COLUMN, PDO::FETCH_GROUP, PDO::FETCH_UNIQUE on fetch(all) calls.

Myself, I don't ever use FETCH_NUM.

> I have the feeling that using different fetch-modes on a single statement/result is a rather exotic use case..

I think it makes it easier to think about that setting default mode, or using setFetchMode simply sets the current default, if one is not provided in query/fetch methods. So the proper fix would be to make reflection know all the types as if it was BOTH internally, but apply the correct type based on current fetch mode.

That being said, I don't think the type of PDOStatement itself matters that much, you want to know the resulting type of query/fetch.

staabm: I have an idea and will try it tomorrow. Stay tuned

tests/defaultFetchAssoc/config/bootstrap.php
```diff
@@ -21,12 +21,13 @@
...
 $config->errorMode(RuntimeConfiguration::ERROR_MODE_EXCEPTION);
 $config->defaultFetchMode(QueryReflector::FETCH_TYPE_ASSOC);
+// $config->debugMode(true);
 
 try {
     if (false !== getenv('GITHUB_ACTION')) {
         $mysqli = @new mysqli('127.0.0.1', 'root', 'root', 'phpstan_dba');
     } else {
-        $mysqli = new mysqli('localhost', 'testuser', 'test', 'phpstan_dba');
+        $mysqli = new mysqli('mysql80.ab', 'testuser', 'test', 'phpstan_dba');
```
staabm, line 30: just realized that you might have different local access parameters in your dev setup @xPaw 

created a issue to support this use-case better:  https://github.com/staabm/phpstan-dba/issues/267

src/QueryReflection/RuntimeConfiguration.php
staabm: lets see whether this works: https://github.com/staabm/phpstan-dba/pull/268

staabm: > That being said, I don't think the type of PDOStatement itself matters that much, you want to know the resulting type of query/fetch.

the type on the statement itself, is the one which is used for the iterator. when its not precise phpstan would think the iterator would contain more offsets, as it actually has.

see 

https://github.com/staabm/phpstan-dba/blob/9a3c964c0497b4dde9204d7e2fd6cf9399c11e05/tests/default/data/pdo.php#L14-L25

staabm: > https://github.com/staabm/phpstan-dba/issues/269

https://github.com/staabm/phpstan-dba/issues/269

PR #257, opened 2022-02-06
extracted `PdoStatementReflection->getStatementResultType()` for re-use

staabm:


src/Extensions/PdoStatementExecuteTypeSpecifyingExtension.php
```diff
@@ -30,7 +30,7 @@ public function getClass(): string
 
     public function isMethodSupported(MethodReflection $methodReflection, MethodCall $node, TypeSpecifierContext $context): bool
     {
-        return 'execute' === $methodReflection->getName();
+        return 'execute' === strtolower($methodReflection->getName());
```
staabm, line 33: drive-by improvement

Issue #255, opened 2022-02-06
Pdo: support `setFetchMode()`


staabm:
```php
<?php
$stmt = $dbh->query('SELECT name, colour, calories FROM fruit');
$stmt->setFetchMode(PDO::FETCH_NUM);
foreach ($stmt as $row) {
    print $row[0] . "\t" . $row[1] . "\t" . $row[2] . "\n";
}
```

https://www.php.net/manual/en/pdostatement.setfetchmode.php

similar to https://github.com/staabm/phpstan-dba/issues/253

PR #258, opened 2022-02-06
PDO: support `PDOStatement::setFetchMode()`

staabm:
closes https://github.com/staabm/phpstan-dba/issues/255


staabm:
hmm maybe those errors are related to the multiple-signautres of [PDOStatement::setFetchMode](https://github.com/phpstan/phpstan-src/blob/master/resources/functionMap.php#L8490-L8493).

I could not reproduce the problems on phpstan.org

staabm:
good point @xPaw - thx.

@ondrejmirtes do you have an idea/a hint, how a type specifying extension could trigger the signature error, which don't occur without the specifying extension?

grafik

ondrejmirtes:
This message says that the type of `$stmt` will be `*NEVER*` after this call. It means that you're narrowing the type to something that's not possible. You can try `$overwrite=true` in the call to `TypeSpecifier::create()` if it makes sense... https://github.com/phpstan/phpstan-src/blob/35606f9ce1e933d4a226e8277176760ac9bd9162/src/Analyser/TypeSpecifier.php#L945-L952

staabm:
> This message says that the type of `$stmt` will be `*NEVER*` after this call

@ondrejmirtes thx for the hint. I did some debugging locally and I can see that I never specify the type of `$methodCall->var` to a `*NEVER*` type.

grafik

I somehow have the feeling that the underlying problem is related to the fact, that `setFetchMode` is defined with multiple signatures in the [resourceMap](https://github.com/phpstan/phpstan-src/blob/72064523053f9b4b317be2e410551c30973968f1/resources/functionMap.php#L8490-L8493).

I have no further idea. do you have another tip, or maybe could even have a closer look at it?

ondrejmirtes:
The type you're specifying it to intersected with the type it already has leads to NeverType.

staabm:
> The type you're specifying it to intersected with the type it already has leads to NeverType.

even if I use "overwrite=true", as I do?

ondrejmirtes:
Not with overwrite, that would be a different problem. Don't be afraid to fire up Xdebug and see what's going on.

staabm:
I guess I tracked it down to:

https://github.com/phpstan/phpstan-src/blob/c5a9c2c50b9b639715ca85541fcfd7f877f5df06/src/Rules/Comparison/ImpossibleCheckTypeHelper.php#L195-L200

where `$isSuperType->no()` is `true` and therefore we run into the `return false` case.
this `return false` triggers the following case

https://github.com/phpstan/phpstan-src/blob/c5a9c2c50b9b639715ca85541fcfd7f877f5df06/src/Rules/Comparison/ImpossibleCheckTypeMethodCallRule.php#L57-L66

----

I think phpstan is wrong as it concludes these 2 `GenericObjectTypes` of the same class with the same generic types are not an supertype of each other

grafik

ondrejmirtes:
Please write a failing test case for that in `GenericObjectTypeTest::dataIsSuperTypeOf()` and we'll see how much sense it makes.

staabm:
just verified the new test works with phpstan@master

Issue #200, opened 2022-01-24
Support for postgresql


stof:
Looking at the code, it seems there is only a MysqliQueryReflector for now. Is Postgresql supported ?

staabm:
correct, atm only mysql is supported

p4veI:
@staabm Hey, I can try to implement this feature into phpstan-dba if you're not against, I'm using pgsql in most of our projects so I'm kind of missing this feature.

staabm:
I would love having pgsql support. we also need to add it to the CI pipeline then

staabm:
@stof @p4veI do you have plans to work on this?

as we recently landed a PDO based reflector in #261 it should now be possible to work out pgsql support.
possible good first steps would be
- [ ] add a github action running pgsql, importing existing db schema and running the testsuite
- [ ] figure out whether pgsql support can be easily built into the PDO based reflector, or whether we better would have 2 separate pdo reflectors (one for mysql, one for pgsql)
- [ ] tackle possible test-failures

this could be done in several pull requests and should not be done in one huge PR.

I am willing to support whoever wants to work on it. As I never used pgsql and don't have any production app using it at hand, I will not work on this topic myself.

p4veI:
@staabm hey, I’m looking forward to looking into this, I didn’t have much time this past week. But I should be able to look into this in the upcoming week..

I already had some base for pgsql actions for github prepared, so I’ll start with that for sure.

PR #261, opened 2022-02-07
Introduce PDO based QueryReflector

p4veI:
Hello, I've tried to setup a PDO based Query reflector to make the extension work with PostgreSQL databases.

I can run the extension in my project now, however I've only worked in some basic support for doctrine 2.*, so I'm thinking if I should try to work the Mapper support now or possibly open up another PR with more improvements and drop the separation from Mysql mapper.

WDYT? is this the way to go?

I'm also wondering about the simulated param value support, I'd love to see some uuid support since I've been experimenting with it a bit.

Following up from https://github.com/staabm/phpstan-dba/issues/200


staabm:
oh wow, this is looking really good already.

I would prefer we could merge the mysqli related refactoring first and do the pgsql stuff separately.

src/QueryReflection/QuerySimulation.php
```diff
@@ -72,13 +75,21 @@ public static function simulateParamValueType(Type $paramType, bool $preparedPar
...
+            || $isStringableObjectType
+            || $isDateTimeType
+        ) {
             // in a prepared context, regular strings are fine
             if (true === $preparedParam) {
                 // returns a string in date-format, so in case the simulated value is used against a date/datetime column
                 // we won't run into a sql error.
                 // XX in case we would have a proper sql parser, we could feed schema-type-dependent default values in case of strings.
-                return date(MysqliQueryReflector::DATE_FORMAT, 0);
+                return $isDateTimeType ? date(self::DATE_FORMAT, 0) : '1';
```
staabm, line 92: moving the DATE_FORMAT into the simulation would only make sense if it works also for PGSQL. does it?

src/QueryReflection/QuerySimulation.php
```diff
@@ -72,13 +75,21 @@ public static function simulateParamValueType(Type $paramType, bool $preparedPar
         }
 
         $stringType = new StringType();
-        if ($stringType->isSuperTypeOf($paramType)->yes() || $paramType instanceof ObjectType && $paramType->isInstanceOf(Stringable::class)->yes()) {
+        $isStringableObjectType = $paramType instanceof ObjectType
+            && $paramType->isInstanceOf(Stringable::class)->yes();
+        $isDateTimeType = $paramType instanceof ObjectType
```
staabm, line 80: I don't think the date-time case you added here works as it is right now.
but lets separate the mysqli refactoring beforehand and discuss this stuff when the PR itself gets smaller and better managable.

src/TypeMapping/PostgresTypeMapper.php
```diff
@@ -0,0 +1,16 @@
...
+namespace staabm\PHPStanDba\TypeMapping;
+
+use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\Type;
+
+final class PostgresTypeMapper implements TypeMapper
+{
+    public function mapToPHPStanType(int $type, int $flags, int $length): Type
+    {
+        throw new ShouldNotHappenException();
```
staabm, line 14: I don't know whether PDO supports this `$flags` and `$type` constants like `mysqli` does... does PDO support this very same concept?

staabm:
another Idea: after the initial refactoring of the existing Mysqli based Reflector landed, we could even build the PDO based Reflector for mysql first and add pgsql as a 3rd step (this might make it easier to work out the PDO specific kinks)

src/TypeMapping/PostgresTypeMapper.php
p4veI: I believe we could get this info with this method https://www.php.net/manual/en/pdostatement.getcolumnmeta.php , but not sure if it's going to work atm.

staabm: We habe this metadata already available in the mysqli type mapper - I guess

staabm:
Landed a few more independent changes out of this PR.

please rebase and add a pgsql based github action

src/QueryReflection/PDOQueryReflector.php
```diff
@@ -0,0 +1,136 @@
...
+use staabm\PHPStanDba\Error;
+use staabm\PHPStanDba\TypeMapping\TypeMapper;
+
+final class PDOQueryReflector implements QueryReflector
+{
+    private const PSQL_INVALID_TEXT_REPRESENTATION = '22P02';
+    private const PSQL_UNDEFINED_COLUMN = '42703';
+    private const PSQL_UNDEFINED_TABLE = '42P01';
+
+    private const SUPPORTED_PDO_CODES = [
```
staabm, line 23:
```suggestion
    private const PDO_ERROR_CODES = [
```

src/TypeMapping/PostgresTypeMapper.php
staabm: Turned differently: in case pdo uses the same flag values, we might even share the type-mapper between the 2 reflector impls

src/QueryReflection/PDOQueryReflector.php
```diff
@@ -0,0 +1,136 @@
...
+    private const MAX_CACHE_SIZE = 50;
+
+    /**
+     * @var array<string, PDOException|array|false>
+     */
+    private array $cache = [];
+
+    public function __construct(private PDO $pdo, private TypeMapper $typeMapper)
+    {
+    }
```
staabm, line 38: Is it required to put pdo into a exception-throwing mode, so it works across php versions consistently?

src/QueryReflection/PDOQueryReflector.php
```diff
@@ -0,0 +1,136 @@
...
+    ];
+
+    private const MAX_CACHE_SIZE = 50;
+
+    /**
+     * @var array<string, PDOException|array|false>
+     */
+    private array $cache = [];
+
+    public function __construct(private PDO $pdo, private TypeMapper $typeMapper)
```
staabm, line 36: the mapper should not be a constructor argument, as these reflector classes will be used within the `phpstan-bootstrap-dba.php` config file and the dba enduser should not be required to pass in classes which he should not know about.

just create and assign the object within the constructor

src/QueryReflection/PDOQueryReflector.php
p4veI: yes that will be required, we can set the attribute for exception throwing to constructor, otherwise it will be left to dba endusers to configure in phpstan-dba-bootstrap

staabm: as phpstan-dba requires the exceptioin fetch mode to work properly, we should set the mode within this reflector

p4veI:
@staabm I tried to push this a little further, tests are still failing but it starting to feel like something :D

I ran the tests with a separate bootstrap (included/switched temporarily), not sure what would be the best way to run tests for different reflectors right now..
I also focus on running PDO with MySQL for now so I didn't go for the psql github action just yet.

Most of the work right now is around the type mapping i think. The tests are failing on mostly integer ranges i think.
Unfortunately `PDOStatement->fetchAll()` doesn't really allow to process anything with `LIMIT 0` queries. So i fiddled with the getColumnMeta method which returns all the info we need for type mapping in the mysqli implementation so it kind of clicked together partially now I think..

Hope this makes sense, let me know.

staabm:
Thanks. I did some example changes so you see how thr CI integration can look like

src/TypeMapping/MysqliTypeMapper.php
```diff
@@ -49,7 +49,7 @@ public function __construct()
         }
     }
 
-    public function mapToPHPStanType(int $mysqlType, int $mysqlFlags, int $length): Type
+    public function mapToPHPStanType(int|string $mysqlType, int|array $mysqlFlags, int $length): Type
```
staabm, line 52: When mysqlFlas is passed in as an the logic within the method will not work?

p4veI: yeah, sorry about this - I just wanted to see what things will be in common for the different mappers..

the logic is similar, but I'm not happy about the union types.. I'm thinking I'll drop the interface for now, or just drop the method definition and have the reflectors dependent of the specific mapper implementation...

staabm: I am fine with dropping the interface, but my intial concerns were more about the duplicated logic of the 2 type-mappers.

but maybe you just need to play a bit longer and figure out, how this can be de-duplicated

staabm:
just finalized the CI setup. I think this should work.. all jobs are running but some error, because of TODOs in the PDO implementation.

I centralized all the db credentials into a `tests/ReflectorFactory.php` file

hopefully you can work with that

src/TypeMapping/MysqlTypeMapper.php
```diff
@@ -23,13 +23,15 @@ final class MysqlTypeMapper
...
+        
         $numeric = false;
         $notNull = false;
         $unsigned = false;
         $autoIncrement = false;
 
         foreach ($mysqlFlags as $flag) {
-            switch ($flag) {
+            switch (strtoupper($flag)) {
                 case 'NUM':
```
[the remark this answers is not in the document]
staabm: I have emulated most of the open issues ... needs still some fine tuning/testing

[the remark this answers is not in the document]
staabm: yeah.. I guess the final task is thinking about how to get similar test coverage for the pdo driver as for the mysqli driver we already have.

src/QueryReflection/PdoQueryReflector.php
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    private function isAutoIncrementCol(string $tableName, string $columnName):bool {
+        if (array_key_exists($tableName, $this->autoIncrementColumns)) {
+            return array_key_exists($columnName, $this->autoIncrementColumns[$tableName]);
+        }
+
+        if ($this->stmt === null) {
+            $this->stmt = $this->pdo->prepare(
+                'SELECT EXTRA FROM information_schema.columns WHERE table_name = ? AND table_schema = DATABASE()'
```
staabm, line 202: We should also detect the UNSIGNED flag in this method

src/QueryReflection/PdoQueryReflector.php
```diff
@@ -0,0 +1,231 @@
...
+            );
+        }
+        $this->stmt->execute([$tableName]);
+        $result = $this->stmt->fetchAll(PDO::FETCH_ASSOC);
+
+        $this->autoIncrementColumns[$tableName] = [];
+        foreach($result as $row) {
+            $extra = $row['EXTRA'] ?? '';
```
staabm, line 210: We might fix the

> Offset 'EXTRA' on array{EXTRA: string} on left side of ?? always exists and is not nullable.

error by using NVL within the query. I guess mariadb has a nullable EXTRA column while mysql has a non-nullable

tests/default/data/pdo.php
```diff
@@ -144,15 +144,6 @@ public function updateQuery(PDO $pdo)
         assertType('PDOStatement', $stmt);
     }
 
-    public function leftJoinQuery(PDO $pdo)
```
staabm, line 147: moved into separate file, because our current way of emulating flags does not support aliases

tests/default/data/pdo.php
```diff
@@ -170,10 +161,10 @@ public function unionParam(PDO $pdo, $adaid, $email)
...
+        assertType('PDOStatement<array{pid: int<0, 4294967295>, c_char5: string, c_varchar255: string, c_varchar25: string|null, c_varbinary255: string, c_varbinary25: string|null, c_date: string|null, c_time: string|null, c_datetime: string|null, c_timestamp: string|null, c_year: int<0, 2155>|null, c_tiny_text: string|null, c_medium_text: string|null, c_text: string|null, c_long_text: string|null, c_enum: string, c_set: string, c_bit: int|null, c_int: int<-2147483648, 2147483647>, c_tinyint: int<-128, 127>, c_smallint: int<-32768, 32767>, c_mediumint: int<-8388608, 8388607>, c_bigint: int, c_double: float, c_real: float, c_boolean: int<-128, 127>, c_blob: string, c_tinyblob: string, c_mediumblog: string, c_longblob: string, c_unsigned_tinyint: int<0, 255>, c_unsigned_int: int<0, 4294967295>, c_unsigned_smallint: int<0, 65535>, c_unsigned_mediumint: int<0, 16777215>, c_unsigned_bigint: int<0, max>}'.$bothType.'>', $stmt);
```
staabm, line 167: adjusted expectations, because they were wrong

src/TypeMapping/MysqlTypeMapper.php
```diff
@@ -83,13 +88,23 @@ public function mapToPHPStanType(?string $mysqlType, array $mysqlFlags, int $len
...
+        // fallbacks
         if (null === $phpstanType) {
-            $phpstanType = match ($mysqlType) {
-                'DOUBLE', 'NEWDECIMAL' => new FloatType(),
+            $phpstanType = match (strtoupper($mysqlType)) {
                 'LONGLONG',
                 'LONG',
                 'SHORT',
-                'YEAR',
+                'TINY',
```
staabm, line 107: TINY was missing in the list. YEAR was taken out and moved before this IF block

staabm:
only thing missing now is separating the Rule* test cases and define different error message expectations for them.

type inference works for all cases but queries containing aliases (I think we can merge without this single test-case).

p4veI:
@staabm Wow, this looks great! Sorry, I didn't have much time to work on this during the week...

staabm:
@p4veI would be great if you could separate the Rules* tests.. at best in a new PR

staabm:
Last remaining todo:

- copy the Rule*MysqliReflectorTest tests to a Rule*PdoReflectorTest counterpart and work out proper test expectations.

staabm:
I think this is a good enough first iteration.

we can reduce code-duplication and try first steps with PGSQL now

Issue #199, opened 2022-01-24
PDO: support bindValue based api


staabm:
example

https://github.com/shopware/shopware/blob/f5146152531467ddd39c7f91b5d7f74f027ca6fc/engine/Shopware/Components/Session/PdoSessionHandler.php?rgh-link-date=2022-01-24T08%3A52%3A08Z#L319-L335

xPaw:
There are also `bindColumn` and `bindParam`.

xPaw:
I tried looking at this, but I can't really wrap my head around php-parser's AST.

It seems that it would be pretty similar to `findQueryStringExpression`, where it finds the parent node, and then traverses the children for all the `bind*` calls.

staabm:
> It seems that it would be pretty similar to `findQueryStringExpression`, where it finds the parent node, and then traverses the children for all the `bind*` calls.

exactly.

staabm:
> I tried looking at this, but I can't really wrap my head around php-parser's AST.

for those cases I put the unit test code into https://phpast.com/ and inspect the AST which gets generated.
it helps to find the proper traversing.

xPaw:
While there's basic support for this now, there's still things to do. In particular with bind calls being inside of `if` statements.

As well as bind calls should be an error when there are also parameters being passed into `execute`.

Not sure if it's worth exploring support for passing in PDO::PARAM_* as third argument in bind calls.

PR #265, opened 2022-02-08
Prototype support for bindValue/bindParam on PDOStatement

xPaw:
I think this adds the right pieces.

refs https://github.com/staabm/phpstan-dba/issues/199

### DIFF
--- a/src/Rules/PdoStatementExecuteMethodRule.php
+++ b/src/Rules/PdoStatementExecuteMethodRule.php
@@ -12,6 +12,9 @@ use PHPStan\Reflection\MethodReflection;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleError;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantIntegerType;
+use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\MixedType;
 use staabm\PHPStanDba\PdoReflection\PdoStatementReflection;
 use staabm\PHPStanDba\QueryReflection\PlaceholderValidation;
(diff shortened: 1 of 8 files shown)


staabm:
This looks great.

we need more testcoverage, especially in
- https://github.com/xPaw/phpstan-dba/blob/pdo-bind-param/tests/default/data/unresolvable-pdo-statement.php
- https://github.com/xPaw/phpstan-dba/blob/pdo-bind-param/tests/default/data/pdo-stmt-execute-error.php

for the error cases.

expected error messages need to be put into
- https://github.com/xPaw/phpstan-dba/blob/pdo-bind-param/tests/default/PdoStatementExecuteMethodRuleTest.php
- https://github.com/xPaw/phpstan-dba/blob/pdo-bind-param/tests/default/UnresolvablePdoStatementRuleTest.php

xPaw:
While adding tests, it seems like it's finding bind calls from previous pdo statements, so the code isn't entirely correct. It needs to stop once it hits `prepare` or something.

staabm:
For inspiration, a few real world examples can be taken from
https://github.com/shopware/shopware/search?q=bindParam

src/Rules/PdoStatementExecuteMethodRule.php
```diff
@@ -71,6 +71,16 @@ private function checkErrors(MethodReflection $methodReflection, MethodCall $met
         $args = $methodCall->getArgs();
         if (\count($args) < 1) {
             $parameters = [];
+
+            $calls = $stmtReflection->findPrepareBindCalls($methodCall);
+
+            foreach ($calls as $bindCall) {
+                $args = $bindCall->getArgs();
+                $parameterName = $args[0]->value->value;
```
staabm, line 79: this needs a `if count ($args) >= 2)` check before accessing `$args[0]` and `$args[1]`

src/Rules/PdoStatementExecuteMethodRule.php
```diff
@@ -71,6 +71,16 @@ private function checkErrors(MethodReflection $methodReflection, MethodCall $met
...
+                $parameterTypes = $scope->getType($args[1]->value);
+
+                $parameters[$parameterName] = $queryReflection->resolveParameters($parameterTypes) ?? [];
```
staabm, line 82: we use the same api a few lines below without this `[$parameterName]` offset, which make me feel that there is a bug somewhere

xPaw: `resolveParameters` accepts a Type which is constructed from the array provided into `execute`, since I don't have such an array of values from bind calls, I have to construct it manually (very possible that I did it incorrectly).

If it's possible to construct some kind of Type array that converts bind calls into an array that was passed into the execute call, it would also work.

staabm: maybe I am mistaken.. came to this conclusion because of the phpstan error in https://github.com/staabm/phpstan-dba/runs/5111061161?check_suite_focus=true of line 98

xPaw: You are right that it currently just puts an empty array there.

xPaw:
I think I fixed it. I manually constructed `ConstantArrayType` out of the collected bind calls.

src/Rules/PdoStatementExecuteMethodRule.php
```diff
@@ -70,16 +73,33 @@ private function checkErrors(MethodReflection $methodReflection, MethodCall $met
...
+            $parameterKeys = [];
+            $parameterValues = [];
+
+            $calls = $stmtReflection->findPrepareBindCalls($methodCall);
+
+            foreach ($calls as $bindCall) {
+                $args = $bindCall->getArgs();
+                if (\count($args) >= 2) {
+                    $keyType = $scope->getType($args[0]->value);
+                    if ($keyType instanceof ConstantIntegerType || $keyType instanceof ConstantStringType) {
```
staabm, line 85: do we need this check here, or could we just feed everything into `$parameterKeys` and `resolveParameters` later on would decide which types are supported and which not?

xPaw: It's constrained by the constructor, so phpstan complained.

staabm (approved):
i love it

tests/default/data/unresolvable-pdo-statement.php
```diff
@@ -38,5 +42,9 @@ public function noErrorOnStringValue(PDO $pdo, string $string)
         $query = 'SELECT adaid FROM ada WHERE email=:email';
         $stmt = $pdo->prepare($query);
         $stmt->execute([':email' => '%|'.$string.'|%']);
+
+        $stmt = $pdo->prepare($query);
+        $stmt->bindValue(':email', '%|'.$string.'|%');
+        $stmt->execute();
```
staabm, line 48: I think we can/should cover a case where some params are bound via `bindValue` or `bindParam` and others are given to `execute` at the same time.. I guess this is supported in pdo?

staabm: would be cool if you could run a small example on your own and verify the thesis.
if its right, we might even create a PHPStan-Rule which errors on `bindValue`/.. calls, when parameters are passed to `execute`

staabm: Cool, thx

https://github.com/staabm/phpstan-dba/issues/275

Issue #254, opened 2022-02-06
PDO: fetch() may return false if there no is no data in exception mode


xPaw:
For code like this:
```php
    $stmt = Container::Database()->query('SELECT AppID FROM Apps LIMIT 5');
    while ($row = $stmt->fetch()) {
        var_dump($row);
    }
```

It will correctly iterate over the rows and it will return `false` once there is no more data.

However `phpstan-dba` assumes the loop never exits:
```
  7      While loop condition is always true.
  11     Unreachable statement - code above always terminates.
```

staabm:
Which fetch mode does you example use?

xPaw:
In my particular code base I use `FETCH_ASSOC` by default, I am not entirely sure whether all fetch modes will behave this way.

xPaw:
That being said, this code currently only supports `PDO::FETCH_ASSOC, PDO::FETCH_NUM, PDO::FETCH_BOTH`, so for all 3 of these it should be fine?

PR #281, opened 2022-02-11
`PDOStatement->fetch()` may return `false` if there no is no data even in exception mode

staabm:
Closes https://github.com/staabm/phpstan-dba/issues/254

`fetchAll()` can return false prior to php8


xPaw:
> fetchAll() can return false prior to php8

Wouldn't it be better to check PHP version and not allow `false` for it?

staabm:
Yep, thats what I wanne do in the end

staabm:
Related phpstan-src PR: https://github.com/phpstan/phpstan-src/pull/1014

Issue #269, opened 2022-02-09
PDO: support more fetch modes


staabm:
- [ ] PDO::FETCH_KEY_PAIR
- [ ] PDO::FETCH_COLUMN
- [ ] PDO::FETCH_GROUP
- [ ] PDO::FETCH_UNIQUE

relevant in
- `$pdo->query($query, $fetchMode);`
- `$stmt->setFetchMode($fetchMode)`
- `$stmt->fetch($fetchMode)`

Seldaek:
PDO::FETCH_CLASS support would also be nice IMO

staabm:
@xPaw since you are using some fetch types we do not yet support type-inference for: do you plan working on the missing fetch-modes?

I will support you in case there are questions left.

xPaw:
I don't currently have plans to look into that, sorry.

xPaw:
One thing worth pointing out is that some of the FETCH constants are actually flags, and not unique modes (thus they can be combined with some modes).

See here:
https://github.com/php/php-src/blob/bf9b774327f139f5770ea87797b933b603f979de/ext/pdo/php_pdo_driver.h#L65-L87

staabm:
> One thing worth pointing out is that some of the FETCH constants are actually flags, and not unique modes (thus they can be combined with some modes).

good point. 

feel free to provide failing unit tests or open issues with concrete examples of things you used/found in the wild.

I am not interessted in adding every theoretical possible combination, but just those things people actually use

PR #302, opened 2022-02-19
PDO: support `PDO::FETCH_COLUMN` in `fetch()`

staabm:

@xPaw could you double check the types are what to expect?
initially I had the feeling this would not be supported, as I was not able to find it in the php.net manual


xPaw:
Seems to be what I would expect

staabm:
thanks for the fast feedbcak

PR #304, opened 2022-02-19
PDO: Added `PDO::FETCH_CLASS` support on `fetch*()`, `fetchObject()`

staabm:

### DIFF
--- a/tests/default/data/pdo-stmt-fetch.php
+++ b/tests/default/data/pdo-stmt-fetch.php
@@ -4,6 +4,7 @@ namespace PdoStmtFetchTest;
 
 use PDO;
 use function PHPStan\Testing\assertType;
+use staabm\PHPStanDba\Tests\Fixture\MyRowClass;
 
 class Foo
 {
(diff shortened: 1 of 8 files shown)


staabm:
@Seldaek could you double check this one matches your expectations?

tests/default/data/pdo-stmt-fetch.php
```diff
@@ -40,6 +41,12 @@ public function fetchAll(PDO $pdo)
         $all = $stmt->fetchAll(PDO::FETCH_KEY_PAIR);
         assertType('array<int, array<string, int<0, 4294967295>>>', $all);
 
+        $all = $stmt->fetchAll(PDO::FETCH_CLASS, MyRowClass::class);
+        assertType('array<int, staabm\PHPStanDba\Tests\Fixture\MyRowClass>', $all);
+
+        $all = $stmt->fetchAll(PDO::FETCH_CLASS);
+        assertType('array<int, stdClass>', $all);
```
Seldaek, line 48: Looks all good to me, but just to make sure, a simple fetch() here would return stdClass|false, right? Might be worth adding for completeness.

staabm: fetch() without args depends on a previous setFetchMode(), which needs https://github.com/staabm/phpstan-dba/pull/258 and which in turn waits for the next phpstan release

Seldaek: Ok I see, great to see it all coming together :)

staabm: The tricky part was figuring out how this stateful apis can work in phpstan context. 

I don't know any extension which did this before, so this is kind of an achievement

PR #306, opened 2022-02-19
Stabilize error/result in cache

### LINKED ISSUE #298: cache item order not stable
sometimes we see diffs between pull requests where the `error` property is changed in the order of the array

grafik

these changes are unncessary noise, as the order of the entries within the array does not matter

xPaw:
Fixes #298

I think there's a different type of cache order issue with `__set_state` which I suspect comes from using newer PHP version (I run php 8.1). All my caches are updated with `keyType` and `itemType` moves above `allArrays`, not really sure what's up with that.

I think `var_export` logic changed in what order it dumps properties?
EDIT: Yep, looks like it changed in PHP 8.1 - https://3v4l.org/Zt4tA

### DIFF
--- a/src/QueryReflection/ReflectionCache.php
+++ b/src/QueryReflection/ReflectionCache.php
@@ -148,9 +148,11 @@ final class ReflectionCache
             // sort records to prevent unnecessary cache invalidation caused by different order of queries
             ksort($newRecords);
 
-            foreach ($newRecords as $newRecord) {
+            foreach ($newRecords as &$newRecord) {
                 ksort($newRecord);
             }
+
+            unset($newRecord);
 
             $cacheContent = '<?php return '.var_export([
                     'schemaVersion' => self::SCHEMA_VERSION,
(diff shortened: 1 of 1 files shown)


staabm:
Thx

staabm:
> I think there's a different type of cache order issue with __set_state which I suspect comes from using newer PHP version (I run php 8.1). All my caches are updated with keyType and itemType moves above allArrays, not really sure what's up with that.

do you mean thats an issue still present - even after merged this PR?

If you know under which circumstances, please open a new issue

xPaw:
> do you mean thats an issue still present - even after merged this PR?

It's a separate issue where the cache will differ when generated under PHP 8.1, I'm not sure you can reasonably fix this because it affects how var_export/serialize orders properties.

See the 3v4l link above, you'll see what I mean (it affects `PHPStan\Type\Constant\ConstantArrayType::__set_state(` in cache.

So this is only an issue when cache is being generated on different versions of PHP.

PR #308, opened 2022-02-19
Use lowercase table name in tests to fix issue on Windows

### LINKED ISSUE #273: Windows: mariadb forces lowercase table names
There's also another unrelated issue with tests as I'm on Windows, and mariadb forces lowercase table names due to case insensitive file system.

Thus one of the tests fails:
```
-41: Query error: Table 'phpstan_dba.unknownTable' doesn't exist (1146).
+41: Query error: Table 'phpstan_dba.unknowntable' doesn't exist (1146).
```

_Originally posted by @xPaw in https://github.com/staabm/phpstan-dba/issues/267#issuecomment-1034751993_

xPaw:
Closes #273

### DIFF
--- a/src/Extensions/PdoStatementFetchDynamicReturnTypeExtension.php
+++ b/src/Extensions/PdoStatementFetchDynamicReturnTypeExtension.php
@@ -130,6 +130,7 @@ final class PdoStatementFetchDynamicReturnTypeExtension implements DynamicMethod
             if (QueryReflector::FETCH_TYPE_KEY_VALUE === $fetchType) {
                 return $rowType;
             }
+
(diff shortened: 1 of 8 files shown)


staabm (approved):
The fix/workaround seems fine for our testsuite. 

I think we will see when phpstan-dba is used in various setups, whether we need a actual fix for the underlying problem.

After re-recording the replay information this can be merged. 

Thank you

xPaw:
I think this might only be a problem for error messages where the table message is printed? Because quering tables in different casing works fine.

staabm:
> I think this might only be a problem for error messages where the table name is printed?

thats one case, correct. I did not think it through whether there are other cases which are relevant.

thank you!

PR #312, opened 2022-02-20
Use transactions when simulating queries

xPaw:
While the code already checks for `SELECT` statements only, this should enforce it on the database level.

For mysqli driver, it will throw "Cannot execute statement in a READ ONLY transaction" error, while in PDO in worse case it will roll back any of the changed data.


staabm:
Thank you

PR #313, opened 2022-02-20
Use phpmyadmin/sql-parser

xPaw:
This is a proof of concept.

One of the 'problems' is if the query does not parse, parser will throw it, instead of sending the supposedly invalid query to the server.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,7 @@
 		"ext-mysqli": "*",
 		"composer-runtime-api": "^2.0",
 		"composer/semver": "^3.2",
+		"phpmyadmin/sql-parser": "^5.5",
 		"phpstan/phpstan": "^1.2"
 	},
 	"require-dev": {
--- a/src/QueryReflection/MysqliQueryReflector.php
+++ b/src/QueryReflection/MysqliQueryReflector.php
@@ -7,6 +7,7 @@ namespace staabm\PHPStanDba\QueryReflection;
 use mysqli;
 use mysqli_result;
 use mysqli_sql_exception;
+use PhpMyAdmin\SqlParser\Parser;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantIntegerType;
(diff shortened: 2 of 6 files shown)


staabm:
I really love the idea of exploring sql parser stuff.

IMO we should not have sql parser types within our interface.
Would it work we hide the sql parser stuff within query simulation?

xPaw:
> Would it work we hide the sql parser stuff within query simulation?

You could do that, but every time you pass around the query it needs to be re-parsed in relevant spots.

staabm:
> but every time you pass around the query it needs to be re-parsed 

would also be interessting whether the parsing is slow .. I guess the parsing will be faster then any sql query we send to the db, so the overheas might be not relevant

xPaw:
Perhaps there should be a Statement interface added? It would contain the original query, have a method to get a simulated version etc.

staabm:
Imo a prototype should at first explore what information we get/need for sql parsing based inference, e.g. 
- can parsing inference be built on top of the existing inference? (Like a 2 phase analysis)
- How fast is the parsing of queries

after we know the answer to this questions we can decide how we should at best integrate the existing api and the sql parser api

xPaw:
Not entirely sure what you mean by 2 phases, wouldn't you want to use the parser to build a proper query for simulation?

I reverted the interface change by the way, it's just parsing in the relevant methods for now.

staabm:
> Not entirely sure what you mean by 2 phases, wouldn't you want to use the parser to build a proper query for simulation?

the already existing features work pretty good without a query parser. we can get better in edge cases, but as is most things I am fine with.

Primarily I would want to use the sql parser to allow type inference in cases, where its not technical possible with the current design (meaning by looking at the result-set types as we do atm).
I don't think we should re-do all the already existing features in a query reflector which understands the query because of query parsing, but instead let one of the already existing reflectors do its magic and after that do a kind of "optimization pass" with the knowledge we win when parsing the simulated query and re-fine with that knowledge the result-set based types.

at least thats the idea how I would do it currently - but I am open for other approaches in case we see a working prototype beforehand

xPaw:
> the PR here - as is - is fine, but solves a "different problem". its mostly a refactoring of already existing featues.

That's what I thought would be a good and simple way to bring the parser in, specifically to get rid of the regex based stuff for query manipulation.

staabm:
> That's what I thought would be a good and simple way to bring the parser in, specifically to get rid of the regex based stuff for query manipulation.

yep thats fine. a "side effect" of having a parser, but not my primary motivation.

xPaw:
sql-parser does parse out expressions, so it is trivial to get the function (like COUNT), I pushed a basic example on how it could work.

staabm:
Cool, thats a first step. 


When thinking about a more involved case like min/max we would need something like

Min(colA, colB)
-> find the phpstan type for colA
-> find the phpstan type for colB
-> combine the 2 types and calculate the resulting phpstan type

The challenge would be on how to get all this information together

PR #318, opened 2022-02-21
PDO: implement `PDO::FETCH_OBJ`

staabm:


tests/default/data/pdo-fetch-types.php
```diff
@@ -23,14 +23,14 @@ public function supportedFetchTypes(PDO $pdo)
 
         $stmt = $pdo->query('SELECT email, adaid FROM ada', PDO::FETCH_BOTH);
         assertType('PDOStatement<array{email: string, 0: string, adaid: int<0, 4294967295>, 1: int<0, 4294967295>}'.$bothType.'>', $stmt);
+
+        $stmt = $pdo->query('SELECT email, adaid FROM ada', PDO::FETCH_OBJ);
+        assertType('PDOStatement<array<int, stdClass>'.$bothType.'>', $stmt);
```
staabm, line 28: phpstan does not allow to define the properties within a `stdClass` (yet?), so we only have this vague structure definition

PR #320, opened 2022-02-24
Add Postgres to Github workflows

p4veI:
Here's some base for postgres setup with phpstan_dba.

I'll have to do more digging regarding the SchemaHasher setup which is now incompatible with psql database. I've included an updated schema for postgres, but I'll have to figure out few more things also. WDYT? Should I add some more stuff into this PR, not sure if perhaps I should disable the hasher for postgres for now so the test case can be run.


src/DbSchema/SchemaHasherMysql.php
```diff
@@ -31,10 +31,17 @@ public function hashDb(): string
             return $this->hash;
         }
 
-        // for a schema with 3.000 columns we need roughly
-        // 70.000 group concat max length
-        $maxConcatQuery = 'SET SESSION group_concat_max_len = 1000000';
-        $this->connection->query($maxConcatQuery);
+        $driver = null;
```
p4veI, line 34: This should perhaps be included in other PR

tests/psql-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+);
+
+ALTER TABLE typemix
+    ADD PRIMARY KEY (pid);
+
+-- CREATE COLLATION german FROM "de_DE";
+CREATE TABLE cmsdomain (
+    id SERIAL,
+    cmsdomainid int NOT NULL,
+--     url varchar(255) COLLATE german NOT NULL,
```
p4veI, line 51: I don't have much experience with collate, this should work but the collation de_DE doesn't exist for UTF8 atm

staabm:
> I'll have to do more digging regarding the SchemaHasher setup which is now incompatible with psql database. I've included an updated schema for postgres, but I'll have to figure out few more things also. WDYT?

the schema hasher is not relevant for a first PR. you can work with `RecordingQueryReflector` for now, which does not require the hasher

tests/psql-schema.sql
```diff
@@ -0,0 +1,59 @@
+CREATE DATABASE phpstan_dba;
```
staabm, line 1: typo in filename: `pgsql-schema.sql`

tests/psql-schema.sql
staabm: I have no idea whether postgres has a similar concept then mysql collations.
maybe we can just turn it into a varchar... collation doesn't matter at all atm for phpstan-dba

tests/psql-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+    cmsdomainid int NOT NULL,
+--     url varchar(255) COLLATE german NOT NULL,
+    standard integer NOT NULL
+);
+
+ALTER TABLE cmsdomain
+    ADD PRIMARY KEY (id);
+
+CREATE INDEX ON cmsdomain(cmsdomainid);
+-- CREATE INDEX ON cmsdomain(url);
```
staabm, line 59: why commented?

.github/workflows/tests-postgres.yml
```diff
@@ -0,0 +1,118 @@
...
+
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - php-version: "8.0"
+                      db-image: 'postgres:14'
+                      reflector: "pdo"
+                      mode: "recording"
+                      dsn: "pgsql"
```
staabm, line 22: why did you choose the term dsn?

In my book a dsn is a string containing db-type, db-host, db-name etc.

do we really need a separate parameter, or maybe we should have a `reflector: pdo-mysql` and a `reflector: pdo-pgsql`?

staabm (changes requested):

tests/ReflectorFactory.php
```diff
@@ -19,11 +19,13 @@ public static function create(string $cacheDir): QueryReflector
     {
         // handle defaults
         if (false !== getenv('GITHUB_ACTION')) {
```
staabm, line 21: I think the cache-file below `$cacheDir.'/.phpunit-phpstan-dba-'.$reflector.'.cache';` needs to be changed, so it contains the database type (we need separate cache files for pgsql)

tests/psql-schema.sql
p4veI: because the collate column was commented, I re-added as varchar with this index

.github/workflows/tests-postgres.yml
staabm: Another point in favor of `pdo-pgsql` is, that we might get a separate `pgsql` reflector based on `pgsql_*` or other apis.

please use `pdo-pgsql` without a new DSN parameter

staabm:
This looks good alread. 

Please focus on implementing the pgsql support within the pdo based reflector for now. 

After we got the pgsql pdo working (on your local machine), I can finish the CI stuff

staabm:
thank you

PR #331, opened 2022-03-05
tests: change adaid type to smallint, to easy cross-db tests

staabm:
changes `ada.adaid` to a smallint type, which should be more compatible with pgsql, which uses the same Integer-Range for `smallint`


tests/schema.sql
```diff
@@ -12,9 +12,6 @@ CREATE TABLE `ada` (
 ALTER TABLE `ada`
     ADD PRIMARY KEY (`adaid`);
 
-ALTER TABLE `ada`
```
staabm, line 15: dropping the auto-increment, because this would change the [type phpstan-dba infers](https://github.com/staabm/phpstan-dba/blob/648405c3ff43999c590a1efa6bf03755a3d117c9/src/TypeMapping/MysqlTypeMapper.php#L87-L89)

PR #335, opened 2022-03-05
Remove ignore-next-line

staabm:
Requires https://github.com/phpstan/phpstan/issues/6743


src/QueryReflection/PdoQueryReflector.php
```diff
@@ -165,33 +165,22 @@ private function simulateQuery(string $queryString)
         $columnIndex = 0;
         while ($columnIndex < $columnCount) {
             // see https://github.com/php/php-src/blob/master/ext/pdo_mysql/mysql_statement.c
+            /** @var ColumnMeta */
```
staabm, line 168:
```suggestion
            /** @var ColumnMeta|false */
```

Issue #278, opened 2022-02-11
Doctrine types support 


Seldaek:
Doctrine supports extra types which can be bound as parameters so it'd be nice if this was supported:

[Doctrine\DBAL\Types\Types](https://github.com/doctrine/dbal/blob/3.3.x/src/Types/Types.php)

One example usage:

```
        $sql = 'UPDATE foo SET last_dt=:last_dt';

        $paramValues = [
            'last_dt' => $lastUsed,
        ];
        $paramTypes = [
            'last_dt' => \Doctrine\DBAL\Types\Type::DATETIME,
        ];

        $updated = $connection->executeUpdate($sql, $paramValues, $paramTypes);
```

Currently this yields:
```
     Internal error: Internal error: Unexpected expression type DateTime of class PHPStan\Type\ObjectType in file /src/Foo.php

     Post the following stack trace to https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md:
     #0 /vendor/staabm/phpstan-dba/src/QueryReflection/QueryReflection.php(258): staabm\PHPStanDba\QueryReflection\QuerySimulation::simulateParamValueType()
     #1 /vendor/staabm/phpstan-dba/src/QueryReflection/QueryReflection.php(223): staabm\PHPStanDba\QueryReflection\QueryReflection->resolveConstantArray()
     #2 /vendor/staabm/phpstan-dba/src/Rules/SyntaxErrorInPreparedStatementMethodRule.php(110): staabm\PHPStanDba\QueryReflection\QueryReflection->resolveParameters()
     #3 /vendor/staabm/phpstan-dba/src/Rules/SyntaxErrorInPreparedStatementMethodRule.php(82): staabm\PHPStanDba\Rules\SyntaxErrorInPreparedStatementMethodRule->checkErrors()
     #4 phar:///vendor/phpstan/phpstan/phpstan.phar/src/Analyser/FileAnalyser.php(102): staabm\PHPStanDba\Rules\SyntaxErrorInPreparedStatementMethodRule->processNode()
```

stof:
Note that you might rely on the phpstan-doctrine tooling to know the conversion between the PHP type and the database type performed by those DBAL types instead of re-inventing your own logic there.
A project using doctrine is likely to have this extension.

PR #339, opened 2022-05-04
Allow simulation of DateTime objects

Seldaek:
Partially addresses #278 - but the way I see it, in QuerySimulation::simulateParamValueType you don't have access to the param types argument, only to the parameter themselves?

I mean `PDO::exec($query, $params, $paramTypes)` <- this param types arg

Without access to that we can serialize the datetime to a Y-m-d H:i:s format, but it won't be always correct, as if you pass Doctrine\DBAL\Types\Types::DATE for ex then it should use Y-m-d only. And strictly speaking we should only simulate/serialize DateTime objects if the appropriate Type is passed in $paramTypes. Otherwise it's unlikely to be a bug if you have a DateTime in there as it will fail at runtime.

This patch does fix the issue I have though so it's a improvement as it allows me to run phpstan-dba on this project of mine, but it's hardly a comprehensive fix.

### DIFF
--- a/tests/default/data/doctrine-dbal.php
+++ b/tests/default/data/doctrine-dbal.php
@@ -222,8 +222,10 @@ class Foo
 
     public function customTypeParameters(Connection $conn)
     {
-        $stmt = $conn->prepare('SELECT count(*) AS c FROM typemix WHERE c_datetime=:last_dt');
-        $result = $stmt->execute(['dt' => new \DateTime()], ['dt' => Types::DATETIME_MUTABLE]); // XXX use bound parameter types instead
+        $result = $conn->executeQuery(
+            'SELECT count(*) AS c FROM typemix WHERE c_datetime=:last_dt',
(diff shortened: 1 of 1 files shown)


staabm:
Could you point me to a code example where this feature is needed?

Seldaek:
There is an example in https://github.com/staabm/phpstan-dba/issues/278 - I don't have anything public sorry. But essentially doctrine allows binding advanced data types, so it'd be great if we could support that here.

staabm:
could you provide a failling unit test?

staabm:
> I mean `PDO::exec($query, $params, $paramTypes)` <- this param types arg

hmm are you sure there is more then 1 arg to `PDO::exec`?

https://www.php.net/manual/en/pdo.exec.php

staabm:
ohh I see, you actually meant `$connection->executeUpdate` and friends

staabm:
@Seldaek just to confirm - this PR as is would unblock you (I am willing to merge it and handle it as a kind of beta-feature)

PR #342, opened 2022-05-18
Add failing tests and TODOs for doctrine types

Seldaek:
Refs #278 and #339

See also doctrine param binding docs: https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#binding-types


tests/default/data/doctrine-dbal.php
```diff
@@ -218,4 +219,49 @@ public function dateParameter(Connection $conn)
...
+
+    public function customTypeParameters(Connection $conn, int $adaid)
+    {
+        $stmt = $conn->prepare('SELECT count(*) AS c FROM typemix WHERE c_datetime=:last_dt');
+        $result = $stmt->execute(['dt' => new \DateTime()], ['dt' => Types::DATETIME_MUTABLE]);
+        assertType('Doctrine\DBAL\Statement<array{c: int, 0: int}>', $result);
+
+        $stmt = $conn->prepare('SELECT count(*) AS c FROM typemix WHERE c_datetime=:last_dt');
+        $result = $stmt->execute(['dt' => new \DateTime()], ['dt' => Types::DATETIME_IMMUTABLE]);
+        // TODO should probably fail as DateTime is passed to a DATETIME_IMMUTABLE type
```
Seldaek, line 231: I don't think doctrine will fail though, the distinction between datetime/datetime_immutable is only made there when deserializing data from DB back into PHP AFAIK, so maybe we can let it go through..

staabm:
thank you!

don't let you mislead by the current build errors.. they seem to be happening also on master - as I realized yesterday.

I need to investigate what needs to be done about them.

Seldaek:
See https://github.com/staabm/phpstan-dba/pull/342/commits/e8b4e6431f3cc297e45e500547b14473072f0262 for a demo of what could be done to solve what I said in #339

> Partially addresses https://github.com/staabm/phpstan-dba/issues/278 - but the way I see it, in QuerySimulation::simulateParamValueType you don't have access to the param types argument, only to the parameter themselves?

>
> I mean PDO::exec($query, $params, $paramTypes) <- this param types arg

>
> Without access to that we can serialize the datetime to a Y-m-d H:i:s format, but it won't be always correct, as if you pass Doctrine\DBAL\Types\Types::DATE for ex then it should use Y-m-d only. And strictly speaking we should only simulate/serialize DateTime objects if the appropriate Type is passed in $paramTypes. Otherwise it's unlikely to be a bug if you have a DateTime in there as it will fail at runtime.

I don't know if this is the best/correct way to do this though, and I'm completely lost as to how to resolve the bound types further, but hopefully it is clear enough that you may be able to go forward with this? Otherwise as I said in #339 just merging that fix (e.g. the first two commits here) would already be an improvement altho it would be very permissive and perhaps not catch mistakes.

staabm:
thank you so far. I will have a closer look when time allows.

staabm:
memo for myself: PDO has [the same bound parameter semantics](https://github.com/staabm/phpstan-dba/pull/339#issuecomment-1130002007)

tests/default/data/doctrine-dbal.php
```diff
@@ -218,4 +219,49 @@ public function dateParameter(Connection $conn)
         $fetchResult = $conn->fetchOne($query, [date('Y-m-d', strtotime('-3hour'))]);
         assertType('int|false', $fetchResult);
     }
+
+    public function customTypeParameters(Connection $conn, int $adaid)
+    {
+        $stmt = $conn->prepare('SELECT count(*) AS c FROM typemix WHERE c_datetime=:last_dt');
+        $result = $stmt->execute(['dt' => new \DateTime()], ['dt' => Types::DATETIME_MUTABLE]);
```
staabm, line 226: does doctrine allow to pass in a `DateTime` object without also giving a parameter-type as a 2nd arg?

or should phpstan-dba error when arg1 is DateTime and arg2 is missing or not equal to `Types::DATETIME_MUTABLE`?

tests/default/data/doctrine-dbal.php
```diff
@@ -218,4 +219,49 @@ public function dateParameter(Connection $conn)
...
+        assertType('Doctrine\DBAL\Result', $result);
+
+        $stmt = $conn->prepare('SELECT count(*) AS c FROM typemix WHERE c_datetime=:last_dt');
+        $result = $stmt->execute(['dt' => new \DateTime()], ['dt' => Types::DATETIME_IMMUTABLE]);
```
staabm, line 230: should this test also use `$conn->executeQuery()` instead? it seems there is no prepare+execute api including bound-parameter-types...?

tests/default/data/doctrine-dbal.php
Seldaek: I am not sure if it lets you do it without the param type, but I'd think not.

tests/default/data/doctrine-dbal.php
Seldaek: DBAL supports this on all connection methods it seems https://github.com/doctrine/dbal/blob/3.3.x/src/Connection.php#L529

Also on quote https://github.com/doctrine/dbal/blob/3.3.x/src/Connection.php#L811

So I guess ideally this would be supported on all the things, but executeStatement/executeQuery would be the must-haves IMO as it's the lowest common denominator and the rest is kinda decorating this.

staabm:
just to let you know. this is not forgotten, but I will do some internal cleanup before starting again on this one

Seldaek:
No worries and no rush really

Issue #340, opened 2022-05-17
Buggy handling of query with FOR UPDATE SKIP LOCKED


Seldaek:
I have this query:

```
SELECT id FROM audit_log WHERE id=:id AND handled=0 LIMIT 1 FOR UPDATE SKIP LOCKED
```

It produces this error:

```
Query error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL/MariaDB server version for the right syntax to use near 'LIMIT 0' at line 1 (1064).
```

If I remove the `FOR UPDATE SKIP LOCKED` it works.

PR #344, opened 2022-05-18
support mysql read concurrency with NOWAIT and SKIP LOCKED

staabm:
see https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html#innodb-locking-reads-nowait-skip-locked

closes https://github.com/staabm/phpstan-dba/issues/340


Seldaek:
Cool thanks!

PR #352, opened 2022-05-19
Move ext-mysqli to require-dev

marmichalski:
Unless there is a special need for mysqli extension to be present, could this be moved to require-dev? Right now, there is also PdoQueryReflector available, but you still have to install mysqli extension to use it.

```sh
$ composer require --dev staabm/phpstan-dba

In PackageDiscoveryTrait.php line 308:

Package staabm/phpstan-dba has requirements incompatible with your PHP version, PHP extensions and Composer version:
  - staabm/phpstan-dba 0.2.26 requires ext-mysqli * but it is not present.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,12 +4,12 @@
 	"license": "MIT",
 	"require": {
 		"php": "^8.0",
-		"ext-mysqli": "*",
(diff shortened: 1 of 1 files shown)


staabm:
Thank you

Issue #357, opened 2022-05-19
PDO Query Reflector, MySQL & json field


marmichalski:
Hi,

`native_type` returned by `getColumnMeta` is missing in case of a JSON column and the query reflector crashes here (because null passed to string arg): https://github.com/staabm/phpstan-dba/blob/9925b3ae41574a2bba71b828252ea0fe46ae25ef/src/QueryReflection/PdoQueryReflector.php#L74

Related: https://bugs.php.net/bug.php?id=78596

Since JSON is returned just as a regular string, the solution might just be to treat it as such?
The dirty solution would be something like this:
```
if ($columnMeta['native_type'] === null && $columnMeta['pdo_type'] === 2) {
    $columnMeta['native_type'] = 'STRING';
}
```

I did not investigate this much, pdo_type 2 has been returned in the meta for json column.

Tested with PHP 8.0.11 and MySQL 5.7.23 (if that matters).

marmichalski:
> Could you create a failling testcase or show a code sample which reproduces your problem?

Of course.

> Which Reflector do you use?

Uhm, the one I've linked, PDO.

marmichalski:
@staabm I've opened a PR; would you like me to work on this or should I leave it with you?

PR #360, opened 2022-05-19
Added support for JSON datatype

marmichalski:
~Expect this to fail (in more places than intended).~

The significant failing job is: [PHPUnit (8.0, mysql:8.0, pdo, recording)](https://github.com/staabm/phpstan-dba/runs/6515449622?check_suite_focus=true), the rest is failing because of cache mismatch (I believe?).

Closes: #357


src/QueryReflection/PdoQueryReflector.php
```diff
@@ -71,6 +71,14 @@ protected function simulateQuery(string $queryString)
                 throw new ShouldNotHappenException('Failed to get column meta for column index '.$columnIndex);
             }
 
+            /*
+             * Native type may not bo set, for example in case of JSON column.
+             * @phpstan-ignore-next-line
+             */
+            if (!isset($columnMeta['native_type'])) {
+                $columnMeta['native_type'] = \PDO::PARAM_INT === $columnMeta['pdo_type'] ? 'INT' : 'STRING';
```
marmichalski, line 79: According to pdo_mysql src, `native_type` can be missing from the return value:
https://github.com/php/php-src/blob/14319c203cb13a7cc0644fb4276d6e62338fda5e/ext/pdo_mysql/mysql_statement.c#L806-L808

and in that case, we may fallback to the `pdo_type`, problem is, according to my understanding of the sources, this can be either INT or STR... https://github.com/php/php-src/blob/14319c203cb13a7cc0644fb4276d6e62338fda5e/ext/pdo_mysql/mysql_statement.c#L811-L827.

src/QueryReflection/PdoQueryReflector.php
```diff
@@ -71,6 +71,14 @@ protected function simulateQuery(string $queryString)
                 throw new ShouldNotHappenException('Failed to get column meta for column index '.$columnIndex);
             }
 
+            /*
+             * Native type may not bo set, for example in case of JSON column.
+             * @phpstan-ignore-next-line
```
marmichalski, line 76: I have not updated the ColumnMeta to represent the actual state of PDO.getColumnMeta return value as just ignoring this line was so much easier... 🤡

staabm:
This looks great, thank you. 

Can the same bug be reproduced with PGSQL?

marmichalski:
> Can the same bug be reproduced with PGSQL?

Kind of. PdoPgQueryReflector does not crash in case of json, because the extension [fallbacks to selecting `TYPNAM` from `PG_TYPE`](https://github.com/php/php-src/blob/46b35c6a500c6ab56064692d45ed3405c1ae8f5a/ext/pdo_pgsql/pgsql_statement.c#L643) and the way phpstan-dba works will make it just return `MixedType`.

This still possibly [may not return `native_type`](https://github.com/php/php-src/blob/46b35c6a500c6ab56064692d45ed3405c1ae8f5a/ext/pdo_pgsql/pgsql_statement.c#L647-L649), but I am unaware of when this could happen, therefore leaving this until reproducible crash scenario is found, which will make it easier to fix 😎

> Additionally it would be cool if you could add a new testcase for the DbaInferenceTest which contains sql queries using the JSON column, to make sure it also works in non trivial `SELECT * FROM x` cases

Yes, I could, but do you have something specific in mind? As `SELECT c_json FROM ...` is probably just the same thing as the `SELECT * ...`? 🤔

staabm:
> Yes, I could, but do you have something specific in mind? As `SELECT c_json FROM ...` is probably just the same thing as the `SELECT * ...`? 🤔

I am thinking about very simple json sql select queries which involve the json columns with e.g. WHERE conditions working on the json column with e.g. json functions.

a few real world statements so we can be sure phpstan-dba does not detect syntax errors etc

staabm:
as this is really useful as is, I will merge for now.

would be great you could add the JSON tests we talked about in a separate PR

PR #362, opened 2022-05-20
cleanup internal types

staabm:


staabm:
hmm not sure yet where this error ist stemming from

PR #364, opened 2022-05-20
Substitute params using regex

### LINKED ISSUE #361: Error when named parameter starts with the name of another named parameter
## Error description
There seems to be an error when resolving named parameters in queries if the parameter name starts with the name of another named parameter.

## Example query
```SQL
SELECT `id`
FROM `table`
WHERE `column` = :param OR `column` = :parameter
```
Note that the **:param**eter param name starts with the exact name of the first named parameter.

 ## Possible cause
 It seems that somewhere the occurences of the first placeholder are replaced by the bound value to that parameter, but it will also replace the first part (`:param`) of the second named parameter inside the query which leads to invalid SQL.

marmichalski:
@staabm should we push the recorded cache with every PR? 🤔 

This PR replaces str_replace with preg_replace + word boundary. If this seems like a dirty solution, the other possibility is to order the parameters by length before substitution I guess.

With str_replace the case added in the syntax-error-in-prepared-statement would fail in `staabm\PHPStanDba\Tests\SyntaxErrorInPreparedStatementMethodRulePdoReflectorTest::testSyntaxErrorInQueryRule` with:
```
+324: Query error: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL/MariaDB server version for the right syntax to use near 'eter LIMIT 0' at line 1 (42000).
```

Closes #361.


staabm:
> should we push the recorded cache with every PR?

yep, as soon as a unit test is added or the impl changes we need to re-record so the testsuite also covers the „record and replay“ case. 

> This PR replaces str_replace with preg_replace + word boundary. If this seems like a dirty solution, the other possibility is to order the parameters by length before substitution I guess.

I am fine with any solution as long as there is no obvious simpler one and it is cocered by a test case

thank you

tests/rules/data/syntax-error-in-prepared-statement.php
```diff
@@ -318,4 +318,9 @@ public function errorOnQueryWithoutArgs(\Doctrine\DBAL\Connection $connection)
         $query = 'SELECT email adaid gesperrt freigabe1u1 FROM ada';
         $connection->executeQuery($query);
     }
+
+    public function preparedNamedParamsSubstitution(Connection $connection)
```
staabm, line 322: I would prefer a separate test file named `bug361.php` So the existing test files will not grow too huge

staabm:
Thank you

PR #369, opened 2022-05-21
added php 7.3 support

staabm:


composer.json
```diff
@@ -3,7 +3,7 @@
 	"type": "phpstan-extension",
 	"license": "MIT",
 	"require": {
-		"php": "^7.4 || ^8.0",
+		"php": "^7.2 || ^8.0",
```
staabm, line 6:
```suggestion
		"php": "^7.3 || ^8.0",
```

.github/workflows/tests.yml
```diff
@@ -16,6 +16,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - php-version: "7.2"
+            db-image: 'mysql:8.0'
+            reflector: "pdo"
```
staabm, line 21:
```suggestion
          - php-version: "7.2"
            db-image: 'mysql:8.0'
            reflector: "pdo"
```

.github/workflows/tests.yml
```diff
@@ -16,6 +16,14 @@ jobs:
...
+            mode: "recording"
```
staabm, line 22:
```suggestion
```

staabm:
@williamdes is this enough for you or do you also need 7.2?

williamdes:
> @williamdes is this enough for you or do you also need 7.2?

To be added to the phpMyAdmin composer json we would need 7.2
I need to test this great tool
I have one question: can we have a config to connect to no database server or something else to provide the necessary data to the tool? 

Because in our CIs there is no available MySQL server

staabm:
if you run phpstan-dba somewhere (e.g. on your local workstation) with a real server you can record the required database interactions and afterwards use this cache files to run the analysis in CI without a active database connection:
https://github.com/staabm/phpstan-dba/blob/main/docs/record-and-replay.md

but tbh, using the tool is easier with a active DB connection, even if you only have the empty database schema in it

williamdes:
> if you run phpstan-dba somewhere (e.g. on your local workstation) with a real server you can record the required database interactions and afterwards use this cache files to run the analysis in CI without a active database connection: https://github.com/staabm/phpstan-dba/blob/main/docs/record-and-replay.md
> 
> but tbh, using the tool is easier with a active DB connection, even if you only have a empty database schema in it

Thank you, I will keep this on my todo list!

staabm:
just give it a try. in case you want to adapt it in phpMyAdmin, I will see what I can do to make php 7.2 support happen

Issue #315, opened 2022-02-20
performance rule: detect queries not using indexes


staabm:

staabm:
Started working on it

PR #377, opened 2022-05-23
mysql query plan analyzer

staabm:
closes https://github.com/staabm/phpstan-dba/issues/315

- analyses the [`EXPLAIN` results](https://dev.mysql.com/doc/refman/8.0/en/explain-output.html#explain_select_type)
- reports sql queries not using an index
- report sql queries doing table-scans
- report sql queries doing too many unindexed reads

ignore rules are applied to reduce the number of false positives:
- ignores [`DERIVED` tables](https://dev.mysql.com/doc/refman/8.0/en/derived-tables.html) because those [cannot get a index](https://github.com/composer/packagist/pull/1304#issuecomment-1137191119)
- ignores missing indices/table-scans for too small tables


src/Analyzer/QueryPlanResult.php
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+
+        return $tables;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getTablesNotEfficient(): array
```
staabm, line 45: needs test

src/Rules/QueryPlanAnalyzerRule.php
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        if (null === $methodReflection) {
+            return [];
+        }
+
+        $unsupportedMethod = true;
+        $queryArgPosition = null;
+        foreach ($this->classMethods as $classMethod) {
+            sscanf($classMethod, '%[^::]::%[^#]#%s', $className, $methodName, $queryArgPosition);
```
staabm, line 67: extract config parsing into a separate class

config/dba.neon
```diff
@@ -56,3 +56,33 @@ services:
             functionNames:
                 - 'Deployer\runMysqlQuery#0'
                 - 'mysqli_query#1'
+
+    -
+        class: staabm\PHPStanDba\Rules\QueryPlanAnalyzerRule
+        tags: [phpstan.rules.rule]
+        arguments:
+            classMethods:
```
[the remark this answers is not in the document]
staabm: will be done in a followup

src/Rules/QueryPlanAnalyzerRule.php
staabm: separate Pr

staabm:
pilot PR in https://github.com/composer/packagist/pull/1304/

src/Analyzer/QueryPlanAnalyzerMysql.php
```diff
@@ -0,0 +1,84 @@
...
+                    $result->addRow($row['table'], QueryPlanResult::NOT_EFFICIENT);
+                }
+                if (\is_int($allowedUnindexedReads) && $row['rows'] > $allowedUnindexedReads) {
+                    $result->addRow($row['table'], QueryPlanResult::NOT_EFFICIENT);
+                }
+            }
+        }
+
+        return $result;
+    }
```
staabm, line 83: @tpetry may I ask you about feedback regarding how to interpret a mysql explain result properly?

I am working on adding perf indication within phpstandba with this PR.
my goal is to emit phpstan errors when queries seem inefficient, indicated by the mysql explain result of said queries.

you can see this feature "in action" in the 'Files Changed'-Tab of https://github.com/composer/packagist/pull/1304/files

example:

grafik

staabm: thx for the feedback. 

my point is that it is helpful when the tool detects queries which don't use a index. the developer might ignore/baseline this error after analyzing the "false-positive". the feature is more about making the developer aware of queries not using indizes - and confirm these.

staabm: I added a threshold, so we only warn about non-indexed and full-table-scan problems for tables with more then 1000 rows

PR #388, opened 2022-05-25
Query Plan Analysis: Reduce default numberOfRowsNotRequiringIndex to 0

staabm:
most of the time, while static analysis there will only be a database schema, no data. therefore we assume that as a default.

in case someone really goes the extra mile and has production quality data `QueryPlanAnalyzer::DEFAULT_SMALL_TABLE_THRESHOLD` should be used to reduce false-positives.


staabm:
Todo: docs anpassen

Issue #385, opened 2022-05-25
support query-plan-analysis on prepared statements


staabm:
atm. this doesn't error:

```
    $stmt = $pdo->query("SELECT * FROM `ada` WHERE email = ?");
    $stmt->execute([$email]); 
```

PR #390, opened 2022-05-27
Fix QueryPlanAnalyser with dynamic offset and limit params

keulinho:
The QueryPlanAnalyser does not strip trailing limits and offsets.
If limits and offsets are set via parameters it would escape the limits and offsets as strings.
This resulted in incorrect SQL being analysed.

### DIFF
--- a/src/QueryReflection/QueryReflection.php
+++ b/src/QueryReflection/QueryReflection.php
@@ -306,7 +306,7 @@ final class QueryReflection
         foreach ($parameters as $placeholderKey => $parameter) {
             $value = $parameter->simulatedValue;
 
-            if (\is_string($value) && !is_numeric($value)) {
+            if (\is_string($value)) {
                 // XXX escaping
                 $value = "'".$value."'";
             } elseif (null === $value) {
--- a/src/QueryReflection/QueryReflection.php
+++ b/src/QueryReflection/QueryReflection.php
@@ -306,7 +306,7 @@ final class QueryReflection
         foreach ($parameters as $placeholderKey => $parameter) {
             $value = $parameter->simulatedValue;
 
-            if (\is_string($value)) {
+            if (\is_string($value) && !is_numeric($value)) {
                 // XXX escaping
                 $value = "'".$value."'";
             } elseif (null === $value) {
--- a/src/QueryReflection/QueryReflection.php
+++ b/src/QueryReflection/QueryReflection.php
@@ -15,6 +15,7 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeUtils;
 use PHPStan\Type\UnionType;
 use staabm\PHPStanDba\Analyzer\QueryPlanAnalyzerMysql;
+use staabm\PHPStanDba\Analyzer\QueryPlanQueryResolver;
 use staabm\PHPStanDba\Analyzer\QueryPlanResult;
 use staabm\PHPStanDba\DbaException;
 use staabm\PHPStanDba\Error;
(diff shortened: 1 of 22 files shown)


keulinho:
I'm not sure about the changes to the cache files here :thinking: i did a `composer record` and committed all changes it produced, not sure if that is the right approach :thinking:

src/QueryReflection/QueryReflection.php
```diff
@@ -306,7 +306,7 @@ private function replaceParameters(string $queryString, array $parameters): stri
         foreach ($parameters as $placeholderKey => $parameter) {
             $value = $parameter->simulatedValue;
 
-            if (\is_string($value)) {
+            if (\is_string($value) && !is_numeric($value)) {
```
keulinho, line 309: not sure if this is the right approach here, as it would not escape numeric values at all now :thinking: 
And at least one test expects that numeric values in the WHERE clause are escaped

Another approach would be to strip the limit and offset of queries here too

tests/rules/data/query-plan-analyzer.php
```diff
@@ -19,7 +19,12 @@ public function noindexDbal(Connection $conn): void
 
     public function noindexPreparedDbal(Connection $conn, string $email): void
     {
-        $conn->executeQuery('SELECT * FROM ada WHERE email = ?', [$email]);
+        $conn->executeQuery('SELECT * FROM ada WHERE email = ?  LIMIT 5 OFFSET 2', [$email]);
```
staabm, line 22: please don't change existing tests.. just add new ones

staabm (commented):
thx for the contribution.

~~I am wondering, as I don't see a testfailure/error, when reverting your fix but just leave the test-changes from this PR.~~

ohh it works now, I see.


> not sure if that is the right approach

thats perfect, thx.

staabm:
I guess the underlying problem is the same as that regular prepared statements do not work properly yet, see https://github.com/staabm/phpstan-dba/issues/385

staabm:
I guess we are still missing some cases, but its already better then before.

Thank you

staabm:
fix released in https://github.com/staabm/phpstan-dba/releases/tag/0.2.34 - thx @keulinho

keulinho:
Thx for your help! :v:

staabm:
@keulinho btw, looks like shopware is using phpstan-dba? do you have any experience to share? feel free to open issues as you find some

keulinho:
Yep we started using this extension last week for shopware 6, because we like the idea of static analysing your DB queries.

But as shopware 6 comes with it own `Data Abstraction Layer` most of the queries are quite abstract and can not be analysed statically (at least not out of the box) there are not so many queries that can be analysed. 

We mostly use plain SQL queries in tests, migrations of performance critical operations, and it is super cool to have this kind of analysis for the latter two cases.

Currently most issues identified were passing unnecessary params when executing an query, but i'm quite curious where this thing is heading too, looks really promising.

staabm:
> But as shopware 6 comes with it own `Data Abstraction Layer` most of the queries are quite abstract and can not be analysed statically (at least not out of the box) there are not so many queries that can be analysed.

if you find examples of this thing, please open a issue. when it e.g. is similar to doctrine dba, it should be easy to re-use existing phpstan-dba functionality for your custom api, see e.g. https://github.com/staabm/phpstan-dba/blob/main/docs/rules.md

keulinho:
> if you find examples of this thing, please open a issue. when it e.g. is similar to doctrine dba, it should be easy to re-use existing phpstan-dba functionality for your custom api, see e.g. https://github.com/staabm/phpstan-dba/blob/main/docs/rules.md

I was not aware that this is possible, looks cool for frameworks that work on SQL basis, but with shopware 6 we tried to build an abstraction around SQL, e.g. you define the data you want to fetch with a `Criteria`-object where you describe which filters should be used and which associations should be loaded etc. And the framework then genereates the SQL for that.

staabm:
> with a `Criteria`-object where you describe which filters should be used and which associations should be loaded etc. And the framework then genereates the SQL for that.

I create phpstan extensions for kind of things in the past, which in turn allows to analyze such dynamic built queries even on phpstan-dba, see https://github.com/staabm/zf-select-strip/blob/a62603941cb2a2aeba6ba121318b93fb93c57430/tests/phpstan/data/zf-select.php#L129-L140

PR #400, opened 2022-05-31
cover unknown constants in tests

staabm:


staabm:
let see we can sent a more meaningful error then what we have right now:

```
phpunit -c tests/rules/config/phpunit.xml
PHPUnit 9.5.20 #StandWithUkraine

..F..SS.SS..SS....                                                18 / 18 (100%)

Time: 00:03.063, Memory: 134.00 MB

There was 1 failure:

1) staabm\PHPStanDba\Tests\QueryPlanAnalyzerRuleTest::testNotUsingIndexInDebugMode
--- Expected
+++ Actual
@@ @@
     💡 see Mysql Docs https://dev.mysql.com/doc/refman/8.0/en/select-optimization.html
 61: Unresolvable Query: Cannot simulate parameter value for type: mixed.
     💡 Make sure all variables involved have a non-mixed type and array-types are specified.
+66: Unresolvable Query: Cannot simulate parameter value for type: *ERROR*.
+    💡 Make sure all variables involved have a non-mixed type and array-types are specified.
 '

phar://C:/dvl/Workspace/phpstan-dba/vendor/phpstan/phpstan/phpstan.phar/src/Testing/RuleTestCase.php:83
C:\dvl\Workspace\phpstan-dba\tests\rules\QueryPlanAnalyzerRuleTest.php:113

```

PR #404, opened 2022-06-09
improve compatibility with phpstan@bleedingEdge

staabm:
see 
- https://github.com/phpstan/phpstan/discussions/7433
- https://phpstan.org/blog/preprocessing-ast-for-custom-rules

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
 		"php": "^7.4 || ^8.0",
 		"composer-runtime-api": "^2.0",
 		"composer/semver": "^3.2",
-		"phpstan/phpstan": "^1.2"
+		"phpstan/phpstan": "^1.5.6"
 	},
 	"require-dev": {
 		"ext-mysqli": "*",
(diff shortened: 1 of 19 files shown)


src/Ast/PreviousConnectingVisitor.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace staabm\PHPStanDba\Ast;
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+use function array_pop;
+use function count;
+use function get_class;
+
+final class PreviousConnectingVisitor extends NodeVisitorAbstract
```
staabm, line 13: this impl is pretty similar to `PhpParser\NodeVisitor\NodeConnectingVisitor` minus the `next` attribute.

I think there is room for improvement, but at first I want to get bleedingEdge compat, so we can play with other bleeding edge features unrelated to memory/performance optimizations

composer.json
```diff
@@ -6,7 +6,7 @@
 		"php": "^7.4 || ^8.0",
 		"composer-runtime-api": "^2.0",
 		"composer/semver": "^3.2",
-		"phpstan/phpstan": "^1.2"
+		"phpstan/phpstan": "^1.5.6"
```
staabm, line 9: we require a higher min version, as described in the linked blog post

PR #413, opened 2022-07-18
implement `@phpstandba-inference-placeholder` annotation

staabm:
in case sql queries need to be built in a way, that phpstan-dba is not able to infer the resulting query string at analysis time, its now possible to typehint methods using a string expression e.g. `@phpstandba-inference-placeholder '1=1'` which will be taken 1:1 into the sql query.

idea is, that there are use-cases in which dynamic query parts are not relevant for the selected colums (so the type inference) and some parts might be ok to ignore regarding syntax error checking.

so you can define a `@phpstandba-inference-placeholder` placeholder, which helps phpstan to understand sql queries at analysis time, which it could not come up with before.


Seldaek:
I don't think I have any of those anymore, but looks like a valuable to have, thanks. Is it possible to do in-line too or do you have to extract the partial logic in a function to be able to use this?

staabm:
thx for the feedback.

at the moments it is only available as annotation on a same-class-method. could you give me an example how a use-case would like where and how you would imagine a in-line example?

staabm:
Related blog post: https://staabm.github.io/2022/07/23/phpstan-dba-inference-placeholder.html

Issue #409, opened 2022-07-06
Support mysqli_execute_query()


staabm:
https://wiki.php.net/rfc/mysqli_execute_query

https://github.com/php/php-src/pull/8660

PR #425, opened 2022-08-22
support mysqli_execute_query (php 8.2+ only)

staabm:
closes https://github.com/staabm/phpstan-dba/issues/409

requires https://github.com/staabm/phpstan-dba/pull/426


staabm:
the test fails because the rule involved runs into this early exit

https://github.com/staabm/phpstan-dba/blob/f60f559183d22df4394f943d7a6a2470e6941fe2/src/Rules/SyntaxErrorInQueryMethodRule.php#L47-L50

I guess it will work with the next phpstan version which has updated php8 stubs

PR #432, opened 2022-10-01
support `@phpstandba-inference-placeholder` on static method calls

staabm:
followup to https://github.com/staabm/phpstan-dba/pull/413/

### DIFF
--- a/src/QueryReflection/QueryReflection.php
+++ b/src/QueryReflection/QueryReflection.php
@@ -191,7 +191,7 @@ final class QueryReflection
         if ($queryExpr instanceof Expr\CallLike) {
             $methodReflection = null;
             if ($queryExpr instanceof Expr\StaticCall) {
-                if ($queryExpr->class instanceof Name) {
+                if ($queryExpr->class instanceof Name && $queryExpr->name instanceof Identifier) {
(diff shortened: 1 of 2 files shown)


src/QueryReflection/QueryReflection.php
```diff
@@ -187,13 +188,21 @@ private function resolveQueryExpr(Expr $queryExpr, Scope $scope): ?string
...
+                if ($queryExpr->class instanceof Name && $queryExpr->name instanceof Identifier) {
+                    $classType = $scope->resolveTypeByName($queryExpr->class);
+                    $methodReflection = $scope->getMethodReflection($classType, $queryExpr->name->name);
+                }
+            } elseif ($queryExpr instanceof Expr\MethodCall && $queryExpr->name instanceof Identifier) {
+                $classReflection = $scope->getClassReflection();
+                if (null !== $classReflection && $classReflection->hasMethod($queryExpr->name->name)) {
+                    $methodReflection = $classReflection->getMethod($queryExpr->name->name, $scope);
+                }
+            }
```
staabm, line 203: we still don't support all calls possible by php, but should be enough for my use-case in REDAXO and `rex_sql::getTablePrefix()` for now

PR #440, opened 2022-10-10
added failling list<int> test

staabm:


staabm:
something is fishy with the test. similar stuff works in https://github.com/staabm/phpstan-dba/pull/688

staabm:
not sure this is doable at all

Issue #442, opened 2022-10-18
Prevent throwing PHPStan internal error when analysing dynamic query with query plans


noemi-salaun:
When using `analyzeQueryPlans` with dynamic query, I get a PHPStan inernal error 

> Internal error: Internal error: Incorrect table name '' in file foo.php

```php
/** @var string $tableExpression */
$snapshotRows = $this->dbal->fetchAllAssociative("SELECT * FROM `$tableExpression`");
```

staabm:
thx for reporting. 

the problem is that the query-inference cannot work with queries which don't contain a table-reference (or a variable table-reference) in FROM.

we should make sure this doesn't error.. but we cannot do type inference/error checking for those cases.

noemi-salaun:
Yes of course, I did not expect the type checking to work on this case, just not to fatal error

staabm:
do you have a full stacktrace for me?

staabm:
I was able to isolate and repro the error on mysqli.

next time please provide a proper backtrace and information about used reflector and phpstandba-bootstrap settings.
at best a reproducing repository or unit test would.

thanks for reporting

noemi-salaun:
To be honest, I thought the error would be easier to deal with. I could have dug deeper and provided more specific information, but you solved the problem too quickly to give me time to react after your last message.

Thank you for your work :heart:

staabm:
no worries. I was kind of suprised, that it was so hard to reproduce ;)

noemi-salaun:
@staabm After updating to v0.2.50 I get a query error even without enabling the debug mode.

>  Query error: Incorrect table name '' (1103).

```php
$rows = $this->dbal->fetchAllAssociative("SELECT * FROM `$tableExpression`");
```

staabm:
@noemi-salaun thx for reporting. fixed with https://github.com/staabm/phpstan-dba/pull/456

PR #447, opened 2022-10-29
fix fatal error on mysql error 1103 / incorrect table

staabm:
closes https://github.com/staabm/phpstan-dba/issues/442

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -87,8 +87,8 @@ jobs:
         # ignore php8.2 requirement error https://github.com/sebastianbergmann/phpunit/issues/5033
         composer-options: "--prefer-dist --no-progress --ignore-platform-req=php+"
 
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -87,8 +87,8 @@ jobs:
         # ignore php8.2 requirement error https://github.com/sebastianbergmann/phpunit/issues/5033
         composer-options: "--prefer-dist --no-progress --ignore-platform-req=php+"
 
-    - name: Setup Problem Matchers for PHPUnit
-      run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+    - name: Configure matchers
```
staabm, line 90: setup-php problem matches does not support custom error messages
https://github.com/shivammathur/setup-php/discussions/451#discussioncomment-683241

Issue #443, opened 2022-10-19
Inconsistent result between run


noemi-salaun:
I can't reproduce it on demand, but it looks like sometime phpstan-dba can infer a query result type, and 5min later, it fails on the same query.

I have a phpstan error ignored in my phpstan.neon :
> '#^Method .* should return array<int, array{nbTransactions: int, platform: string}> but returns array<int, array<string, mixed>>\.$#'

After installing phpstan-dba and running it successfuly, nothing change for this error.

One day later, my CI fail telling me that this ignored error isn't triggered anymore. So I tried running phpstan locally and yes, phpstan know the proper type for this query. So I remove the ignored error from my phpstan.neon, commit everything, it works.

Then 5min later, my phpstan warn me about the exact same error I just remove from ignored because it wasn't necessary anymore...

I tried removing the phpstan and phpstan-dba entirely but it keeps throwing the same error and I cannot make it detect the correct return type. The generated cache for phpstan-dba seems to vary between calls :/

staabm:
the latest phpstan release fixed some result-cache issues.
https://github.com/phpstan/phpstan/releases/tag/1.8.11

could you re-validate whether this might have fixed the underlying problem?

noemi-salaun:
I think the inconsistency came from phpstorm which autorun phpstan on the current file, so phpstan-dba cache only get info about this file and its dependencies, instead of the whole project.

But that does not explain why sometime phpstan-dba successfuly get my query return type, and sometime not.
For example, my code below :

```php
$dbal = $this->getEntityManager()->getConnection();

$sql = <<<SQL
SELECT COUNT(bt.id) AS nbTransactions, bm.key AS platform
FROM BankTransaction bt
    JOIN BankTransactionStatus bts ON bt.statusId = bts.id
    JOIN BankMethod bm ON bt.methodId = bm.id
WHERE userId = :userId
    AND bts.status = 'paymentback_done'
GROUP BY bm.id, bm.key
SQL;

return $dbal->fetchAllAssociative($sql, ['userId' => $userId]);
```

If I run phpstan on my whole project, I got an empty result cache, but if I run it on this specific file, or its direct parent directory it works fine.

<details>
  <summary>CLICK TO EXPAND Empty result cache when running phpstan on whole project</summary>

```php
<?php return array (
  'schemaVersion' => 'v9-put-null-when-valid',
  'schemaHash' => 'b7adec1533cb7674c8fc90d6d83e8804',
  'records' => 
  array (
[...]

Issue #310, opened 2022-02-20
cache files grow endless with `RecordingQueryReflector`


staabm:
as we are only adding and overriding informtion in the cache-files created by `RecordingQueryReflector` the cache file contents grow over time.

the cache files can contain stale information, as long as the very same query is re-evaluated and therefore the corresponding information is updated.

cache informations of queries which the application under test might no longer trigger will stay in the cache files.

it would be great if we could make sure stale information within the cache files would be deleted when queries are no longer relevant.

staabm:
since we now have `ReplayAndRecordingQueryReflector` I think this should be less relevant, as the cache files will be invalidated on every db schema change

PR #454, opened 2022-11-02
don't overwrite reflection cache, when analyzing only a single file in phpstorm

staabm:
references https://github.com/staabm/phpstan-dba/issues/443#issuecomment-1292041749

related phpstan discussion https://github.com/phpstan/phpstan/discussions/8259


staabm:
@noemi-salaun could you please give this change a try. I think it should prevent phpstorm-dba to write/change the result cache, when the phpstan analysis was started from within phpstorn

noemi-salaun:
@staabm It looks good. If I update a query, the new query is added in the cache file but everything else is left untouched in the cache file :+1: 
If I keep modifying the same request, every change add a new query in the cache file. Obsolete queries are keeped in the cache file even when I run phpstan manually. I need to remove the cache and re-run phpstan to get a clean cache without obsolete queries.

staabm:
> Obsolete queries are keeped in the cache file even when I run phpstan manually. I need to remove the cache and re-run phpstan to get a clean cache without obsolete queries.

yeah, thats a known issue, see https://github.com/staabm/phpstan-dba/issues/310
it shouldn't be a problem though, when obsolete queries are contained in the cache file

staabm:
I can no longer reproduce the underlying issue. we need a failing unit test first

Issue #462, opened 2022-11-12
Nette/dibi driver support


jakubvojacek:
Hello

I tried modifying phpstan-dba to add support to either https://github.com/dg/dibi or https://github.com/nette/database (both are very very similar in terms of methods, almost interchangeable).

The syntax is mainly

```php
$connection = new \Dibi\Connection([...]);
// $connection->getDriver()->getResource() returns instance of \mysqli
$connectoin->query('SELECT * from users where users_name = ?', $username)->fetch();
```

i tried copying stuff from mysqli/pdo driver but they are different, they execute the query in 2 steps (prepare, execute) while dibi/nette does that in one method that accepts N-parameters.

Could you please guide me a bit how to add basic support for this? 

Thanks a lot
Jakub

staabm:
Hi Jakub,

great to see people interessting in adding support for new db apis.

to add a new api to phpstan-dba you need 2 things:
1) adding the query-string taking methods to the syntax error detection rule by adding full qualitfied method in
https://github.com/staabm/phpstan-dba/blob/0f46ee86370633d8a69eaf5cdb5319af496df2ed/config/dba.neon#L15-L91

2) adding type inference via PHPStan ReturnTypeExtensions.
you can take inspiration from [already existing extensions](https://github.com/staabm/phpstan-dba/tree/main/src/Extensions). I am not at all familiar with any of the proposed apis, but I think they look similar to the already existing Doctrine DBAL stuff (which also separates the connection and the result object).
see 
- https://github.com/staabm/phpstan-dba/blob/0f46ee86370633d8a69eaf5cdb5319af496df2ed/src/Extensions/DoctrineConnectionQueryDynamicReturnTypeExtension.php
- https://github.com/staabm/phpstan-dba/blob/0f46ee86370633d8a69eaf5cdb5319af496df2ed/src/Extensions/DoctrineResultDynamicReturnTypeExtension.php

I would suggest doing these 2 things in separate PRs. starting with 1) which is the easier step.

feel free to open a PR with your attemps, so I can give  you more concrete hints

jakubvojacek:
Well 1) seems easy (I already kind of did that when I added those methods in services in the neon file) when I was testing, afaik the it would be this

```diff
🐞  jakubvojacek@~/Sites/phpstan-dba $ git diff
diff --git a/config/dba.neon b/config/dba.neon
index 2542a3d..250bdda 100644
--- a/config/dba.neon
+++ b/config/dba.neon
[...]

PR #463, opened 2022-11-13
Detect syntax errors in `Dibi\Connection::*`

jakubvojacek:
Hello @staabm 

as discussed in https://github.com/staabm/phpstan-dba/issues/462, I am attempting to add support for [dibi](https://github.com/dg/dibi) database driver.

So far I have come up with those changes. Surprisingly, it kinda works. Already got some reports such as:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Customer.php                                                                                                                                                                                                       
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  193    Parameter #1 $key of function array_key_exists expects int|string, int<0, 4294967295>|null given.                                                                                                                  
  246    Parameter #2 $data of method SomeApp\SchemaProcessor\SchemaProcessor::processOutput() expects array<string, mixed>, array<int|string, array<int|SomeApp\ApiModule\Entity\ProfileEntity>|int|string|null> given.  
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

It correctly guesses the parameters and their types. 

But,

1) the part where I replace the question mark with `"1"` seems messy. I guess you got some APIs already prepared to replace placeholders? But I couldn't really find it.
2) It does not report syntax errors correctly because of the hack where I replace placeholders in inferType instead of the APIs handling it I think. It should be somehow replaced in `resolveQueryStrings` automatically right?
3) phpcs/phpstan fails, hope that is okay for now.


config/dba.neon
```diff
@@ -33,6 +33,7 @@ services:
                 - 'Doctrine\DBAL\Connection::iterateAssociativeIndexed'
                 - 'Doctrine\DBAL\Connection::iterateColumn'
                 - 'Doctrine\DBAL\Connection::executeUpdate' # deprecated in doctrine
+                - 'Dibi\Connection::fetch'
```
staabm, line 36: please rebase your branch and move the configs into `rules.neon`.

also please add test-cases into `SyntaxErrorInPreparedStatementMethodRuleTest`, see
- `syntax-error-in-prepared-statement.php`
- `syntax-error-with-inference-placeholder.php`

which shows that the rule is detecting syntax errors in queries given to `Dibi\Connection::fetch`.
also test a case which shows that queries without errors don't get errors reported.

config/dba.neon
```diff
@@ -49,6 +50,7 @@ services:
                 - 'mysqli::execute_query#0'
                 - 'Doctrine\DBAL\Connection::query#0' # deprecated in doctrine
                 - 'Doctrine\DBAL\Connection::exec#0' # deprecated in doctrine
+                - 'Dibi\Connection::fetch#0'
```
staabm, line 53: I guess since you added this method already in `SyntaxErrorInPreparedStatementMethodRule` we don't need it here again. you can verify this theses after you added test coverage and drop this line afterwards

config/extensions.neon
```diff
@@ -90,3 +90,7 @@ services:
         class: staabm\PHPStanDba\Ast\PreviousConnectingVisitor
         tags:
         	- phpstan.parser.richParserNodeVisitor
+    -
+        class: staabm\PHPStanDba\Extensions\DibiConnectionFetchDynamicReturnTypeExtension
```
staabm, line 94: please separate the phpstan extensions (and reflection) related stuff into a separate PR.
we will come back to them after we landed syntax error detection first.

staabm (commented):
thanks for the PR. please find my feedback in the comments.

staabm:
btw: `QueryPlanAnalyzerMysql` is a advanced feature, which is not required to merge this very first PR for bidi

config/dba.neon
jakubvojacek: > please rebase your branch and move the configs into the newly added rules.neon.

done

> also please add test-cases into SyntaxErrorInPreparedStatementMethodRuleTest, see

not really sure how to add tests there. There's comparison for the reflector, eg `if (MysqliQueryReflector::NAME === getenv('DBA_REFLECTOR')) {` but you said that DibiQueryReflector shouldnt be part of this PR. Can you please explain further?

config/dba.neon
jakubvojacek: removed

config/extensions.neon
jakubvojacek: removed

jakubvojacek:
I am not planning to do anything with `QueryPlanAnalyzerMysql` but since I change the type of `$connection`, I had to changed the docblocks there too

config/dba.neon
staabm: if I got it right, Dibi supports a mysql backend?
I did not check whether it also supports postgres.
these are the 2 only backend phpstan-dba atm supports

if I am right, for the syntax query check the dibi method will just use one of the existing mysql backend connectors. we don't need to run the queries thru bidi itself at analysis time.

jakubvojacek: > if I got it right, Dibi supports a mysql backend?

that is correct. Nevertheless, I am still confused as to what to do with the tests :(

staabm: I will add a single test so you get an idea.

tests/rules/data/syntax-error-in-prepared-statement.php
```diff
@@ -328,4 +328,13 @@ public function bug442(Connection $conn, string $table)
...
         $conn->executeQuery("SELECT * FROM `$table`");
     }
+
+    public function bidiTest(\Dibi\Connection $conn)
+    {
+        // syntax error, should expect a error in the test
+        $conn->fetch('SELECT email adaid WHERE gesperrt FROM ada');
+        // all good, no error expected
+        $conn->fetch('SELECT email, adaid FROM ada');
+    }
```
staabm, line 338: with the single line configuration change in `rule.neon` phpstan-dba is now able to detect this syntax error on line 335 (for mysql and postgres adapter).

I think at this point you can look into dibi and add more tests of methods which take a sql query as its arg (while also adding new entries in `rules.neon`.

staabm: btw: looking at the dibi docs, I found examples like
> $database->query('UPDATE users SET', $arr, 'WHERE id=?', $id);

I don't think we can easily add support for these notation, where sql construction and values are intermixed

jakubvojacek: I added the methods. The thing is that it gets complicated with the `query` method. 

`$connection->query()` returns an object on which you can call `fetch, fetchAll, fetchSingle, fetchPairs`. Or you can use the shortcut for each of these (`$connection->fetch(query)` instead of `$connection->query(query)->fetch()`)

Not really sure if this should be addressed or we should just force to use the shorter version (w/o `query()`)

> btw: looking at the dibi docs, I found examples like

yes, I was thinking that could be problematic. Perhaps this advanced syntax would not be supported at first, start with the basics

staabm: > `$connection->query()` returns an object on which you can call `fetch, fetchAll, fetchSingle, fetchPairs`. Or you can use the shortcut for each of these (`$connection->fetch(query)` instead of `$connection->query(query)->fetch()`)

lets have a look in https://github.com/dg/dibi/blob/1585255d7d99869c313f054e1b389b45a4f5d678/src/Dibi/Connection.php#L216

we see that this method returns a object from type `Dibi\Result`. This class does not contain methods for which you can pass in a query-string, so for the goal of this PR, thats not relevant.

last step for this PR before merge: add a example similar to 
`$database->query('UPDATE users SET', $arr, 'WHERE id=?', $id);`
so we can see that `SyntaxErrorInPreparedStatementMethodRule`  does not freak out / fatal erros or creates false positives in such scenarios.

after that, this PR is good to be merged and I can give you new directions on how we can make the type inference happen

jakubvojacek: I have added one test

staabm:
could you fix the remaining expectation difference in the pgsql unit test?

see https://github.com/staabm/phpstan-dba/actions/runs/3460165478/jobs/5776403740

```
1) staabm\PHPStanDba\Tests\SyntaxErrorInPreparedStatementMethodRuleTest::testSyntaxErrorInQueryRule
--- Expected
+++ Actual
@@ @@
                            ^ (42601).
 335: Query error: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "FROM"
 LINE 1: SELECT email adaid WHERE gesperrt FROM ada LIMIT 0
-                           ^ (42601).
+                                          ^ (42601).
 336: Query error: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "FROM"
 LINE 1: SELECT email adaid WHERE gesperrt FROM ada LIMIT 0
-                           ^ (42601).
+                                          ^ (42601).
 337: Query error: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "FROM"
 LINE 1: SELECT email adaid WHERE gesperrt FROM ada LIMIT 0
-                           ^ (42601).
+                                          ^ (42601).
 338: Query error: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "FROM"
 LINE 1: SELECT email adaid WHERE gesperrt FROM ada LIMIT 0
-                           ^ (42601).
+                                          ^ (42601).
 339: Query error: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "FROM"
 LINE 1: SELECT email adaid WHERE gesperrt FROM ada LIMIT 0
-                           ^ (42601).
+                                          ^ (42601).
 '
 ```

jakubvojacek:
> could you fix the remaining expectation difference in the pgsql unit test?

all checks passed

staabm:
awesome, thank you.

so next stop is type inference.
I got the impression that the `Dibi->fetch` connection works pretty similar to the `DoctrineConnection->fetch`.
therefore I would start with a new dibi extension inspired by `DoctrineConnectionFetchDynamicReturnTypeExtension`.

please search for `doctrine-dbal.php` within the repo to get an idea how we can test type inference for bidi and create a new `bidi.php` or similar, with some initial expectations

staabm:
@jakubvojacek I had to push out a new release for PHPStan 1.9.3 compat. therefore the initial batch of Dibi support is available in the stable release for initial testing

staabm:
would be great you could add a few words about configuration for Dibi in the readme

jakubvojacek:
Hello 

yes, I'd be happy to but not before https://github.com/staabm/phpstan-dba/pull/481 is resolved since that would block the majority of users from a good experience

PR #465, opened 2022-11-15
Dibi phase two

jakubvojacek:
I had to revert some of the changes from the last commit. The syntax in between dibi and mysqli is so different that I decided to create new `rule` `SyntaxErrorInDibiPreparedStatementMethodRule` (mainly due to how parameters and placeholders were treated)

I also added driver `dibi-mysql` into tests (but not sure whether they will be executed, pipelines are not running in my fork). The tests for other rules are now failing on `Unsupported DBA_REFLECTOR` when I manually used `DBA_REFLECTOR` environment value. 

I wasnt sure whether it should be resolved tho, why should the test cases that have nothing to do with `dibi-mysql` support it.

Please let me know what do you think. I used this on our codebase which has around 100k lines and it worked (it actually helped us find a few issues that we overlooked).

Also, there is one phpstan rule failing, the `staabm\PHPStanDba\Error` seems kinda strict when it comes to `$code`. In my reflector it's just some integer (I dont really care what the code was, it failed and the message displayed by phpstan is clear enough), 

Thanks
Jakub

### DIFF
--- a/src/DbSchema/SchemaHasherMysql.php
+++ b/src/DbSchema/SchemaHasherMysql.php
@@ -11,7 +11,7 @@ use PHPStan\ShouldNotHappenException;
 final class SchemaHasherMysql
 {
     /**
-     * @var PDO|mysqli
+     * @var \Dibi\Connection|PDO|mysqli
      */
     private $connection;
 
@@ -21,7 +21,7 @@ final class SchemaHasherMysql
     private $hash = null;
 
     /**
-     * @param PDO|mysqli $connection
+     * @param \Dibi\Connection|PDO|mysqli $connection
      */
     public function __construct($connection)
     {
@@ -62,6 +62,11 @@ final class SchemaHasherMysql
         $hash = '';
         if ($this->connection instanceof PDO) {
             $stmt = $this->connection->query($query);
+            foreach ($stmt as $row) {
+                $hash = $row['dbsignature'] ?? '';
+            }
+        } elseif ($this->connection instanceof \Dibi\Connection) {
+            $stmt = $this->connection->query($query)->fetchAll();
             foreach ($stmt as $row) {
                 $hash = $row['dbsignature'] ?? '';
             }
--- a/src/QueryReflection/DibiQueryReflector.php
+++ b/src/QueryReflection/DibiQueryReflector.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+namespace staabm\PHPStanDba\QueryReflection;
+
+use Dibi\Connection;
+use Dibi\ConstraintViolationException;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\Type;
+use staabm\PHPStanDba\Error;
+use staabm\PHPStanDba\TypeMapping\MysqliTypeMapper;
+
+final class DibiQueryReflector implements QueryReflector, RecordingReflector
+{
+    public const NAME = 'dibi-mysql';
+
+    /** @var array<string, \Throwable|list<object>|null> */
+    private $cache = [];
+
+    /** @var Connection */
+    private $connection;
+
+    /**
+     * @var MysqliTypeMapper
+     */
+    private $typeMapper;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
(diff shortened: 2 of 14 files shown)


tests/default/DbaInferenceTest.php
```diff
@@ -19,6 +19,8 @@ public function dataFileAsserts(): iterable
 
         if ('pdo-pgsql' === getenv('DBA_REFLECTOR')) {
             yield from $this->gatherAssertTypes(__DIR__.'/data/pdo-pgsql.php');
+        } elseif ('dibi-mysql' === getenv('DBA_REFLECTOR')) {
+            yield from $this->gatherAssertTypes(__DIR__.'/data/dibi-mysql.php');
```
staabm, line 23: I think you can move the ` yield from $this->gatherAssertTypes(__DIR__.'/data/dibi-mysql.php');` to the bottom of this method.

dibi will re-use the pdo-pgsql/mysqli etc reflector.. we don't have a `dibi-mysql` reflector. you can drop the if condition

staabm:
> I had to revert some of the changes from the last commit. The syntax in between dibi and mysqli is so different that I decided to create new rule SyntaxErrorInDibiPreparedStatementMethodRule (mainly due to how parameters and placeholders were treated)

can you show me a code example of this problem?

tests/default/DbaInferenceTest.php
jakubvojacek: the PR has it afaik https://github.com/jakubvojacek/phpstan-dba/blob/dibi-phase-two/tests/ReflectorFactory.php#L70

src/DbSchema/SchemaHasherMysql.php
```diff
@@ -65,6 +65,11 @@ public function hashDb(): string
             foreach ($stmt as $row) {
                 $hash = $row['dbsignature'] ?? '';
             }
+        } elseif ($this->connection instanceof \Dibi\Connection) {
```
staabm, line 68: I am wondering, why we have this dibi connection here.
does dibi support a way to get the underlying connection out of a dibi connection, so we can handle it like any other mysql/pdo/mysqli connection?

src/QueryReflection/DibiQueryReflector.php
```diff
@@ -0,0 +1,126 @@
...
+use Dibi\Connection;
+use Dibi\ConstraintViolationException;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\Type;
+use staabm\PHPStanDba\Error;
+use staabm\PHPStanDba\TypeMapping\MysqliTypeMapper;
+
+final class DibiQueryReflector implements QueryReflector, RecordingReflector
```
staabm, line 16: I don't think we need this reflector, in case we could get the underlying connection out of a dibi-connection

jakubvojacek:
the placeholders are completely different. It doesnt have named (`:smt`) and unnamed (`?`) placeholders. Instead it has got `%i, %s, %t, %d, %and, %or, %lmt, %ex, %sql, ...`. 

Also, the existing rule was only using argv[1] and assuming it was an array with variables. The dibi method have `1...n` parameters

src/QueryReflection/DibiQueryReflector.php
jakubvojacek: we can (via `$connection->getDriver()->getResource()`) but what about the way I replace placeholders in `DibiQueryReflector::simulateQuery`? Also the `fetchtypes` parameter in `getResultType` is irrelevant for dibi (not even sure what it is)

src/DbSchema/SchemaHasherMysql.php
jakubvojacek: changed in latest commit to use `$connection->getDriver()->getResource()`

staabm:
> I used this on our codebase which has around 100k lines and it worked (it actually helped us find a few issues that we overlooked).

thats great news. congratulations.

I will have a closer look at the PR when time allows

jakubvojacek:
Since both PRs require `DibiQueryReflector` - do you agree with keeping it? Thats why I'm doing it in one slightly bigger PR. At least I was able to test it all together unlike in the previously merged PR (during inference development i came across things that werent possible without dibiqueryreflector (at least in my opinion)).

Would it be a big problem to keep it in one PR - to make sure that both parts run smoothly?

staabm:
ok, if these are dependent lets keep it for now. I will have a look at it and come back to you.

PR #469, opened 2022-11-16
Fix inference in dynamic return extension

jakubvojacek:
I could push into your PR so I create a new one.

is this what you mean by fixing the extension? The tests passed when i tried to run them locally afaik. Hopefully they will pass CI here as well


staabm:
thanks. one small glitch left it seems.

PR #470, opened 2022-11-17
Fix dibi tests and on placeholder in inference test

jakubvojacek:


staabm:
good morning :-)

please rebase.. hopefully after I merged https://github.com/staabm/phpstan-dba/pull/471 you get better error messages here

jakubvojacek:
There is still one failing pdo-pgsql test which starts to fail when I add placeholders to dibi.php

Doesnt work:
```php
$row = $connection->fetch('SELECT email, adaid FROM ada where email = %s', 'email@github.com');
```

Works:
```php
$row = $connection->fetch('SELECT email, adaid FROM ada where email = %s', 'email@github.com');
```

I am not sure tho why this test should pass. Dibi requires mysql not pgsql, its only logical that it wont pass because the simulateQuery that is in PdoPgSqlQueryReflector wont resolve the placeholders correctly?

Or am I missing here something?

staabm:
I think dibi supports pgsql

grafik


> I am not sure tho why this test should pass. Dibi requires mysql not pgsql, its only logical that it wont pass because the simulateQuery that is in PdoPgSqlQueryReflector wont resolve the placeholders correctly?

IIRC the latest PR has changes which moved the placeholder resolving into the dibi rule/extension, so it should work independently of the reflector

jakubvojacek:
I installed it locally and found the problem. It was double vs single quotes (pgsql requires single quotes only unlike mysql)

jakubvojacek:
I havent ever used the record & replay feature (I always got database accessible). Can you please give me some guidance @staabm as to how to simulate these errors locally? I tried 

```
export DBA_REFLECTOR='pdo-mysql'
export DBA_MODE='recording'
phpunit -c tests/default/config/phpunit.xml '--colors=always'
export DBA_MODE='replay-and-recording'
phpunit -c tests/default/config/phpunit.xml '--colors=always'
```

and its all passing, no errors

I guess some cache isnt being populated but I havent touched that part in my PR. That should be done in the mysql-reflector somewhere I guess

staabm:
when your local setup works you need to run `composer record` and then commit the cache files beeing created by this process

jakubvojacek:
finally :) next i'd like to resolve https://github.com/staabm/phpstan-dba/issues/472

staabm:
awesome, thank you

PR #475, opened 2022-11-18
Dibi\Row return type for dibi

jakubvojacek:
```php
        assertType('int<-32768, 32767>', $row->adaid);
        assertType('int<-32768, 32767>', $row['adaid']);
```

those assertions were failing locally when i was running tests

```diff
--- Expected
+++ Actual
@@ @@
-'int<-32768, 32767>'
+'*ERROR*'

Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'int<-32768, 32767>'
+'mixed'
```

### DIFF
--- a/config/DibiRow.stub
+++ b/config/DibiRow.stub
@@ -3,12 +3,13 @@
 namespace Dibi;
 
 /**
- * @template rowType
+ * @template T
  *
- * @implements \Traversable<int, rowType>
- * @implements \IteratorAggregate<int, rowType>
+ * @implements \Countable<T>
+ * @implements \IteratorAggregate<int, T>
+ * @implements \ArrayAccess<string, T>
  */
-class Row implements \Traversable, \IteratorAggregate
+class Row implements \ArrayAccess, \IteratorAggregate, \Countable
 {
 
 }
--- a/config/DibiRow.stub
+++ b/config/DibiRow.stub
@@ -3,11 +3,10 @@
 namespace Dibi;
 
 /**
- * @template T
- *
- * @implements \Countable<T>
- * @implements \IteratorAggregate<int, T>
- * @implements \ArrayAccess<string, T>
+ * @template TKey of array-key
+ * @template TValue
+ * @implements \IteratorAggregate<TKey, TValue>
+ * @implements \ArrayAccess<TKey, TValue>
  */
 class Row implements \ArrayAccess, \IteratorAggregate, \Countable
 {
--- a/config/DibiRow.stub
+++ b/config/DibiRow.stub
@@ -1,0 +1,14 @@
+<?php
+
+namespace Dibi;
+
+/**
+ * @template rowType
+ *
+ * @implements \Traversable<int, rowType>
+ * @implements \IteratorAggregate<int, rowType>
+ */
+class Row implements \Traversable, \IteratorAggregate
+{
+
+}
--- a/tests/default/data/dibi.php
+++ b/tests/default/data/dibi.php
@@ -11,7 +11,8 @@ class Foo
     {
         $row = $connection->fetch('SELECT email, adaid FROM ada where adaid = %i', 1);
         assertType('Dibi\Row<array{email: string, adaid: int<-32768, 32767>}>|null', $row);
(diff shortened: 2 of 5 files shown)


config/DibiRow.stub
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Dibi;
+
+/**
+ * @template rowType
+ *
+ * @implements \Traversable<int, rowType>
+ * @implements \IteratorAggregate<int, rowType>
+ */
+class Row implements \Traversable, \IteratorAggregate
```
staabm, line 11: per https://github.com/dg/dibi/blob/cab1c5b5e69f9a6c745919145b06ce82f291a7c5/src/Dibi/Row.php#L17

this should be `class Row implements \ArrayAccess, \IteratorAggregate, \Countable` (and you need to adjust the phpdocs acordingly)

tests/default/data/dibi.php
```diff
@@ -10,13 +10,15 @@ class Foo
     public function fetch(\Dibi\Connection $connection)
     {
         $row = $connection->fetch('SELECT email, adaid FROM ada where adaid = %i', 1);
-        assertType('array{email: string, adaid: int<-32768, 32767>}|null', $row);
+        assertType('Dibi\Row<array{email: string, adaid: int<-32768, 32767>}>|null', $row);
+        assertType('int<-32768, 32767>', $row->adaid);
```
staabm, line 14: I think to make phpstan aware of the "magic properties" of `Row` we need an additional `PropertiesClassReflectionExtension` for `Row`

https://phpstan.org/developing-extensions/class-reflection-extensions

tests/default/data/dibi.php
```diff
@@ -10,13 +10,15 @@ class Foo
...
+        assertType('int<-32768, 32767>', $row['adaid']);
```
staabm, line 15: the array access should be covered by the stubs generic type and `\ArrayAccess` with said type

config/DibiRow.stub
jakubvojacek: I updated the class declaration but im still lost with the phpdocs as to what should be there :(

staabm: lookup those class definitions in [phpstan-src stubs](https://github.com/phpstan/phpstan-src/tree/1.9.x/stubs)
there you can see which class has which generics defined and which don't have generics at all.

jakubvojacek: this is where I was looking already (i got there from https://phpstan.org/user-guide/stub-files) but this is the first stub file that I am writing :(

I updated the doc block by copying definitions from arrayaccess (https://github.com/phpstan/phpstan-src/blob/1.5.x/stubs/ArrayObject.stub) but still no luck

```php
/**
 * @template TKey of array-key
 * @template TValue
 * @implements \IteratorAggregate<TKey, TValue>
 * @implements \ArrayAccess<TKey, TValue>
 */
```

staabm:
Will be on vacation for a few days. I will come back to you next week

jakubvojacek:
thanks :+1: I tried extending the stub-ed Dibi\Row with ArrayObject (phpstan wanted me to remove the implemets arrayAccess annotations - i guses because the arrayobject already implements it).

Anyhow, I am completely lost with this stub :(

config/DibiRow.stub
staabm: I played arround a bit with this PR. I think stubs won't help us for the problem at hand.

just opened a discussion at https://github.com/phpstan/phpstan/discussions/8399 to brainstorm how we can get this polymorphic object represented

jakubvojacek: Thanks, I appreciate greatly the effort :+1:

staabm: I have no clue whether its the way to go, but you might try the suggestion here: https://github.com/phpstan/phpstan/discussions/8399#discussioncomment-4190224

jakubvojacek: `methodReflectionExtension`, I will give it a try, thanks

staabm: maybe its even https://phpstan.org/developing-extensions/class-reflection-extensions#properties-class-reflection-extensions

Issue #476, opened 2022-11-18
Dibi - date(time) columns return type


jakubvojacek:
Hello @staabm 

the dibi driver returns date/datetime/timestamp columns as \DateTimeImmutable object.

```diff
diff --git a/vendor/staabm/phpstan-dba/src/TypeMapping/MysqlTypeMapper.php b/vendor/staabm/phpstan-dba/src/TypeMapping/MysqlTypeMapper.php
index 43d929b70..548ba4927 100644
--- a/vendor/staabm/phpstan-dba/src/TypeMapping/MysqlTypeMapper.php
+++ b/vendor/staabm/phpstan-dba/src/TypeMapping/MysqlTypeMapper.php
@@ -10,10 +10,12 @@ use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\MixedType;
+use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
 use staabm\PHPStanDba\QueryReflection\QueryReflection;
 use staabm\PHPStanDba\Types\MysqlIntegerRanges;
 
@@ -144,11 +146,13 @@ final class MysqlTypeMapper implements TypeMapper
                 case 'STRING':
                 case 'VAR_STRING':
                 case 'JSON':
+                    $phpstanType = new StringType();
+                    break;
                 case 'DATE':
                 case 'TIME':
                 case 'DATETIME':
                 case 'TIMESTAMP':
-                    $phpstanType = new StringType();
+                    $phpstanType = new ObjectType(\DateTimeImmutable::class);
                     break;
                 default:
                     $phpstanType = new MixedType();
```

I could make it work properly by these changes. 

How do you think we could integrate this into existing codebase?

I was thinking i could pass third parameter into `public function getResultType(string $queryString, int $fetchType): ?Type`, eg `?string $calledFromClass = null`

and call it from DibiConnectionFetchReturnDynamicType with `self::class` parameter

If you agree, I will open a PR

staabm:
Since dibi supports more then just mysql we need to come up with something db independent. 

I imagine other dbal layers having similar magic for other types as well. 

I have to think about it .. suggestions welcome.

jakubvojacek:
I will create PR with the logic described above for you to look at

[...]

PR #477, opened 2022-11-19
Datetime support

jakubvojacek:
related to https://github.com/staabm/phpstan-dba/issues/476


staabm:
I think we should add a `DbalApi` class like

```
final class DbalApi {
  public const API_DIBI = 'dibi';
  public const API_DOCTRINE_DBAL = 'doctrine-dbal';
  //.. 

  /** @param self::API_* $type
  public function __construct($api) {}
  
  public function returnsDateTime():bool {
    return $this->api === self::API_DIBI;
  }
}
```

and pass it to the `QueryReflection` constructor.

in https://github.com/staabm/phpstan-dba/blob/c28ca1759967d3433f31493bdba05c09db332ab4/src/QueryReflection/QueryReflection.php#L373 we can then pass it further into the reflector (I think we might need a new method in the`QueryReflector` interface)

jakubvojacek:
> we can then pass it further into the reflector (I think we might need a new method in theQueryReflector interface)

its an interface, I cannot add method to query reflector. Just the signature and then i'd have to add the method body to all 7 seven reflectors that implement `QueryReflection` and inside the QueryReflection I'd have to pass it to constructor of the Mapper class.

Should I go ahead and do that? Also, do you think the dbapi should be nullable?  so that it would only be set when some special return would be required (now only for dibi)?

```diff
--- a/src/QueryReflection/QueryReflection.php
+++ b/src/QueryReflection/QueryReflection.php
@@ -43,6 +43,16 @@ final class QueryReflection
      */
     private static $runtimeConfiguration;
 
+    /**
+     * @var DbaApi|null
+     */
+    private $dbaApi;
+
+    public function __construct(?DbaApi $dbaApi = null)
+    {
+        $this->dbaApi = $dbaApi;
+    }
+
`

staabm:
> Also, do you think the dbapi should be nullable?

yes, I think so.



> its an interface, I cannot add method to query reflector.

feel free to add a method. I don't expect anyone is implementing it outside of phpstan-dba.

jakubvojacek:
ok, on it

PR #479, opened 2022-11-20
feat: pgsql query plan analyzer support

p4veI:
Hi, it took me a while to get to this, but I finally managed to take a look at this and I'm keen for some comments regarding the features.

I might have to do a bit more digging regarding postgres indexes, as I currently think the differences are quite big and frankly I'm not sure at this point if this doesn't require a bit different approach - regarding generating tips/generating errors etc.

It's kind of hard for me to figure out if the behaviour I've built in now is somewhat similar to what the mysql analyzer is doing or not, I'll probably discuss with friends in the upcoming week.

So it's kind of a draft for now..

I also noticed pgsql cache is now ignored, what's the reason for that (I guess it complicates maintenance for building mysql features)?


tests/rules/QueryPlanAnalyzerRuleTest.php
```diff
@@ -60,7 +57,9 @@ public function testNotUsingIndex(): void
         $this->numberOfRowsNotRequiringIndex = 2;
 
         $proposal = "\n\nConsider optimizing the query.\nIn some cases this is not a problem and this error should be ignored.";
-        $tip = 'see Mysql Docs https://dev.mysql.com/doc/refman/8.0/en/select-optimization.html';
+        $tip = 'pdo-pgsql' === getenv('DBA_REFLECTOR')
+            ? QueryPlanResult::PGSQL_TIP
+            : 'see Mysql Docs https://dev.mysql.com/doc/refman/8.0/en/select-optimization.html';
```
staabm, line 62: I think we should also put the mysql tip onto `QueryPlanResult` then

staabm:
Hey!

nice to see you again here :-).

this looks pretty solid to me. I don't have much to add.

> I might have to do a bit more digging regarding postgres indexes, as I currently think the differences are quite big and frankly I'm not sure at this point if this doesn't require a bit different approach - regarding generating tips/generating errors etc.

I have 0 experience with pgsql, therefore cannot judge whether this is the way to go. would be great you could double check that with your friends.

> I also noticed pgsql cache is now ignored, what's the reason for that (I guess it complicates maintenance for building mysql features)?

I guess you are talking about test coverage for the record and replay functionality? 

I don't have pgsql running locally and therefore rely on CI mostly for this stuff. in case we can cover more, without requiring myself to install pgsql locally is very welcome.

thank you

src/Analyzer/QueryPlanResult.php
```diff
@@ -6,12 +6,31 @@
 
 final class QueryPlanResult
 {
-    public const NO_INDEX = 'no-index';
-    public const TABLE_SCAN = 'table-scan';
-    public const UNINDEXED_READS = 'unindexed-reads';
+    public const ROW_NO_INDEX = 'no-index';
```
staabm, line 9: renaming existing constants is a BC break. Please leave the old ones additionally in and mark them as deprecated

PR #480, opened 2022-11-20
implement DbaApi

jakubvojacek:
implements changes suggested in https://github.com/staabm/phpstan-dba/pull/477


src/QueryReflection/BasePdoQueryReflector.php
```diff
@@ -136,6 +136,15 @@ public function getResultType(string $queryString, int $fetchType): ?Type
         return $arrayBuilder->getArray();
     }
 
+    public function setupDbaApi(?DbaApi $dbaApi): void
```
staabm, line 139: I think we should leave the method in the base class but implement it in the subclasses, so we don't need this ugly if-condition which couples the base class to the actual subclasses

src/QueryReflection/ChainedReflector.php
```diff
@@ -61,4 +61,8 @@ public function getResultType(string $queryString, int $fetchType): ?Type
 
         return null;
     }
+
+    public function setupDbaApi(?DbaApi $dbaApi): void
+    {
+    }
```
staabm, line 67: I think this should be

```
foreach ($this->reflectors as $reflector) {
  $reflector->setupDbaApi($dbaApi);
}
```

src/QueryReflection/DbaApi.php
```diff
@@ -0,0 +1,24 @@
...
+    /** @var self::API_* */
+    private $api;
+
+    /** @param self::API_* $api */
+    public function __construct($api)
+    {
+        $this->api = $api;
+    }
+
+    public function returnsDateTime(): bool
```
staabm, line 20:
```suggestion
    public function returnsDateTimeImmutable(): bool
```

src/QueryReflection/LazyQueryReflector.php
```diff
@@ -43,4 +43,8 @@ public function getResultType(string $queryString, int $fetchType): ?Type
 
         return $this->reflector->getResultType($queryString, $fetchType);
     }
+
+    public function setupDbaApi(?DbaApi $dbaApi): void
+    {
+    }
```
staabm, line 49: store the `dbaApi` in a local property and pass it via `setupDbaApi` after reflector creation.

might be worth it to extract the reflector creation into a new private method

staabm (changes requested):
this looks great, just noted a few nits.

src/QueryReflection/ChainedReflector.php
jakubvojacek: hopefully done

src/QueryReflection/DbaApi.php
jakubvojacek: hopefully done

src/QueryReflection/LazyQueryReflector.php
jakubvojacek: > store the dbaApi in a local property and pass it via setupDbaApi after reflector creation.

done

> might be worth it to extract the reflector creation into a new private method

wasnt exactly sure what do you mean

staabm: `$this->reflector = ($this->reflectorFactory)();` can be found 2 times in this class.

I think we should extract a method

```
private function createReflector():QueryReflector {
    if (null === $this->reflector) {
            $this->reflector = ($this->reflectorFactory)();
            $this->reflector->setupDbaApi($this->dbaApi);
    }
    return $this->reflector;
}
```

and use it from this 2 call sites

jakubvojacek: Sorry, I overlooked the 2nd usage :( Hopefully done

staabm (commented):
looks great, thx.

could you please adjust all remaining `new QueryReflection()` spots in the codebase so we have a consistent use of `DbaApi` within this repo?

I think we only need a few new constants, like `DbaAPI::API_DOCTRINE_DBAL`, `DbaAPI::API_PDO`, `DbaAPI::API_MYSQLI`.
after that we are good to go.

src/QueryReflection/LazyQueryReflector.php
staabm: no worries. thx for working on it.

staabm:
ok lets give it a try as is then. thank you

PR #481, opened 2022-11-22
[WIP] Dibi return type 2

jakubvojacek:
Hello @staabm 

I tried to implement property extension to handle `$row->xxx` cases.

It actually almost works.

I just dont know how to get the property that phpstan-dba guessed in the `DibiPropertyClassReflectionExtension` on line 30. I need to pass whatever was set in the DibiReturn extension for given propery.

Could you please take a look at it to see whether you got some ideas?

I though that `$classReflection->getProperty($name, $scope)` looks the best from the list of methods that IDE suggested but I couldnt find out where to get the scope parameter from

Thanks

### DIFF
--- a/src/Extensions/DibiPropertyClassReflectionExtension.php
+++ b/src/Extensions/DibiPropertyClassReflectionExtension.php
@@ -1,0 +1,116 @@
+<?php
+
+declare(strict_types=1);
+
+namespace staabm\PHPStanDba\Extensions;
+
+use Dibi\Row;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\PropertiesClassReflectionExtension;
+use PHPStan\Reflection\PropertyReflection;
+use PHPStan\TrinaryLogic;
+use PHPStan\Type\IntegerType;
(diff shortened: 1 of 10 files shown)


src/Extensions/DibiPropertyClassReflectionExtension.php
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        return false;
+    }
+
+    public function getProperty(ClassReflection $classReflection, string $propertyName): PropertyReflection
+    {
+        // how to get the type of property from classReflection?
+
+        return new DibiRowPropertyReflection($classReflection, new IntegerType());
```
staabm, line 30: we can't infer the type from the ClassReflection as the types are depedent on the GenericObjectType, not on the class level

jakubvojacek: I tried using the stub files like suggest by Ondrej but couldnt make it work (while it works in the sandbox https://phpstan.org/r/4adcc760-1f94-45bb-a1ef-ac25c82a030b, it doesnt work in the PR)

Issue #394, opened 2022-05-30
phpstan-dba not picking up array shape on doctrine/dbal result


Brammm:
I ran into a PHPStan error on the following code: https://phpstan.org/r/0a244b72-7cea-4746-8d69-26b17871fd62

After installing this package, I was still getting the same error, but noticed the cache file also wasn't being generated, so I figured something must not be getting recognized. After poking around a little bit, I changed my `fetchData` method to use the connection to immediately fetch:

```php
private function fetchData(AccessTokenEntityInterface $accessTokenEntity): array
{
    $data = $this->connection->fetchAssociative(
        <<<SQL
        SELECT
            u.id,
            u.name,
            u.email,
            u.timezone,
            u.language,
            u.company_id
        FROM user u
        WHERE u.email = :email
        SQL,
        [
            'email' => 'john@example.com', // in actual code, this is calling a method on $accessTokenEntity
        ]
    );

    if (! $data) {
        throw new UserNotFoundException();
    }

    return $data;
}
```

Now I did notice the cache file got generated, but I'm still getting the original error. 

```php
<?php return array (
  'schemaVersion' => 'v9-put-null-when-valid',
  'schemaHash' => '56eb246db59f605bc0f4aa5c9242e8bb',
  'records' => 
  array (
    'SELECT
    u.id,
    u.name,
    u.email,
    u.timezone,
    u.language,
    u.company_id
FROM user u
WHERE u.email = \'1970-01-01\'' => 
    array (
      'error' => NULL,
    ),
  ),
  'runtimeConfig' => 
  array (
    'errorMode' => 'default',
    'debugMode' => false,
    'stringifyTypes' => false,
  ),
);
```

I'm guessing this package does not get triggered when fetching from the Statement result?

Secondly, am I doing something wrong that my DB isn't getting properly analyzed? I find it a little strange that a date is being used as the example email.

staabm:
thanks for the report.

since you are running phpstan-dba with a custom api it seems (not PDO, Doctrine, mysqli or similar), you need to configure the rules for your API, see https://github.com/staabm/phpstan-dba/blob/main/docs/rules.md

> Secondly, am I doing something wrong that my DB isn't getting properly analyzed? I find it a little strange that a date is being used as the example email.

how is your database schema looking for the table involved?

Brammm:
Just to confirm, I am using `doctrine/dbal`. 

This is my bootstrap file:

```php
<?php

[...]

PR #482, opened 2022-11-23
test with empty result cache

staabm:
refs https://github.com/staabm/phpstan-dba/issues/394#issuecomment-1312118095


staabm:
@noemi-salaun I think this should fix your problem mentioned in https://github.com/staabm/phpstan-dba/issues/394#issuecomment-1312118095

src/QueryReflection/ReflectionCache.php
```diff
@@ -334,14 +331,7 @@ public function putResultType(string $queryString, int $fetchType, ?Type $result
             $this->changes[$queryString]['result'][$fetchType] = $this->records[$queryString]['result'][$fetchType] = $resultType;
             $this->cacheIsDirty = true;
         }
-    }
-
-    private function isPHPStormSingleFileRun(): bool
```
staabm, line 339: dropping the phpstorm workaround again. the workaround also made the terminal integration not persisting the cache anymore (no matter fullproject scan or not)

Issue #472, opened 2022-11-17
Support for checking update/insert/delete statemenet


jakubvojacek:
Hello

in my dibi fork, I was able to successfully test update/delete/insert statements (inside a transaction). @staabm commented it out as not supported yet (https://github.com/staabm/phpstan-dba/pull/468).

Could it be supported since it worked fine in my fork? We'd have to make sure we disallow any statement that does work inside the transaction (eg put whitelist only to select, insert, update, delete to disallow truncate, alter etc)

Thanks

staabm:
since we are running in transactios for a few releases without further problem, I am fine with enabling writable queries.

we should make sure though, that we have enough test coverage (also for other extensions, not just dibi)

staabm:
Just to make sure we are on the same page. This feature request is about syntax validation, not about type inference?

jakubvojacek:
> This feature request is about syntax validation,

yes, I plan to work on it next week most likely

hemberger:
Hi! Sorry to intrude, but I was wondering if there are plans to support convenience wrapper methods once this work is completed, such as the `update()`, `insert()`, and `delete()` methods of `Doctrine\DBAL\Connection`. I saw a discussion about `QueryBuilder` in https://github.com/staabm/phpstan-dba/issues/94, but it seems like these methods might be significantly more straightforward.

Thanks for all your work on this. It's quite exciting!

staabm:
please open a separate issue with a concrete code example and your expectations.

staabm:
implemented with https://github.com/staabm/phpstan-dba/pull/483

staabm:
@hemberger @jakubvojacek released with https://github.com/staabm/phpstan-dba/releases/tag/0.2.55

the feature is opt-in and should work for all supported apis (mysqli, doctrine, dibi, pdo) and all supported backends (mysql, mariadb, pgsql)

please give it a try and report back

jakubvojacek:
Works like a charm, it actualyl found one issue with one insert query that I wasnt aware of :+1:

PR #483, opened 2022-11-25
Support for delete, insert, update, replace

jakubvojacek:
related to https://github.com/staabm/phpstan-dba/issues/472#issuecomment-1327085284

> add test-coverage / expectations for error checking and also type inference

the new statements are pretty much the same as SELECT - the logic regarding placeholders is the same. Do we need to test something more since the features are already being tested for SELECT statements?

> widen the whitelist of statement types in places like

whitelisted for SELECT, INSERT, DELETE, UPDATE

> make sure we did not miss a place where queries are executed against the database and don't have a transaction yet

I checked mysql/pdo-mysql/pdo-pf reflectors and transactions were used


jakubvojacek:
I got a problem with the tests - in order for them to succeed, they need to succeed locally (becuase of recoridng & replay). But the output differs here and in CI

```
ID`, `changet' at line

vs

ID`, `chan...' at line
```

the 3 remaining failures seems to be caused by this. I could manually modify the recorded files but that wouldnt be easy to do after every commit.

```

File: tests/rules/data/syntax-error-in-prepared-statement.php [hunk 1]
```diff
@@ -322,6 +322,7 @@ public function errorOnQueryWithoutArgs(\Doctrine\DBAL\Connection $connection)
     public function preparedNamedParamsSubstitution(Connection $connection)
     {
         $connection->preparedQuery('SELECT email FROM ada WHERE email = :param OR email = :parameter', ['param' => 'abc', 'parameter' => 'def']);
+        $connection->preparedQuery('INSERT into ada(adaid, gesperrt, email,freigabe1u1) values(:adaid, :gesperrt, :email, :freigabe1u1)', ['adaid' => 1, 'gesperrt' => 0, 'email' => 'test@github.com', 0]);
```
staabm, line 325: I think the unit test should verify that the insert/update/deletes did not actually modify data in the tables (the statements are rolled back)

tests/rules/data/syntax-error-in-dibi-prepared-statement.php
```diff
@@ -50,7 +50,6 @@ public function testIgnoredPlaceholders(\Dibi\Connection $conn)
         $conn->query('SELECT %n, %n frommmm %n', 'email', 'dadid', 'ada');
     }
 
-    /* phpstan-dba does not yet support writable queries
     public function testDeleteUpdateInsert(\Dibi\Connection $conn)
```
staabm, line 53:
```suggestion
    public function testInvalidDeleteUpdateInsert(\Dibi\Connection $conn)
```

src/QueryReflection/QueryReflection.php
```diff
@@ -56,7 +56,12 @@ public static function setupReflector(QueryReflector $reflector, RuntimeConfigur
 
     public function validateQueryString(string $queryString): ?Error
     {
-        if ('SELECT' !== $this->getQueryType($queryString)) {
+        if (!\in_array($this->getQueryType($queryString), [
+            'SELECT',
+            'INSERT',
+            'DELETE',
+            'UPDATE',
```
staabm, line 63: what about `REPLACE`?

src/QueryReflection/QuerySimulation.php
```diff
@@ -129,7 +129,10 @@ public static function simulate(string $queryString): ?string
         if (null === $queryString) {
             return null;
         }
-        $queryString .= ' LIMIT 0';
+
+        if ('SELECT' === QueryReflection::getQueryType($queryString)) {
```
staabm, line 133:
```suggestion
        // make sure we don't unnecessarily transfer data, as we are only interessted in the statement is succeeding
        if ('SELECT' === QueryReflection::getQueryType($queryString)) {
```

tests/rules/data/syntax-error-in-prepared-statement.php
staabm: does the current logic error when passing e.g. a `float` to a parameter expecting a `string`?
we should test that

jakubvojacek: > I think we need a unit test to verify that the insert/update/deletes did not actually modify data in the tables (the statements are rolled back)

not sure where in tests I can get the actual connection - the test cases there are using phpstan rules/return type DI and I wasnt sure how to get the actual connection from there (I was thinking of using QueryReflection::reflector() but thats private)

```php
    public function testInsertRollback()
    {
        $countBefore = $connection->fetchSingle('SELECT count(*) FROM ada');
        assertNotNull($countBefore);

        $connection->query('INSERT into ada(adaid, gesperrt, email, freigabe1u1) values(4, 1, "test@github.com", 0)');

        $countAfter = $connection->fetchSingle('SELECT count(*) FROM ada');
        assertNotNull($countAfter);
        assertEquals($countAfter, $countBefore);
    }
```

I had in mind something like this but these asserts didnt work when I added this code into dibi.php

src/QueryReflection/QueryReflection.php
jakubvojacek: added

tests/rules/data/syntax-error-in-prepared-statement.php
staabm: How the test works dependa on the Testclass beeing used. 

DbaInferenceTestCase is for type inference- therefore you use assertType() in the files beeing included.

Other testcases test only a single rule and you need to assert, within the test-class itself, at which line errors are expected for the files beeing included.

staabm: Maybe we can just test expectating, whether start transaction and rollback was invoked by the test beeing executed?

jakubvojacek: > You might even add a new plain phpunit testcase, which executes the query and tests expectations

but i dont think I can get the db driver from reflector, its all private. and even if i did have the connection in a new plain phpunit test case, how could I test the reflector? Since the reflection/type inference is tested inside the extended phpstan testcases

> Maybe we can just test expectating, whether start transaction and rollback was invoked by the test beeing executed?

you mean to track what commands were executed on the driver and check that there was start transaction / rollback? But normal asserts do not work inside phpstan testcases (i tried adding `assertNotNull(1);` in dibi.php and the tests passed)

src/QueryReflection/QuerySimulation.php
staabm: I think we can add the `LIMIT` to all queries except `INSERT`.
it also makes sense to limit `UPDATE`, `REPLACE` and `DELETE`.. that way we are double save to not modify real data (even though we have the transaction doing the same for us... but better save then sorry)

jakubvojacek: done

staabm: removed non-SLECT types again as they did not work for non-mysql

staabm:
thank you

PR #506, opened 2023-02-13
Allow placeholder validation for write queries

### LINKED ISSUE #503: Write queries with missing parameters fail silently
I think I've isolated an issue where queries with errors are not being reported. If a write query is missing a parameter, then it seems that no errors are reported for that query under any circumstances.

I'm using a custom prepared statement method (using PDO under the hood) with the following configuration:
```neon
class: staabm\PHPStanDba\Rules\SyntaxErrorInPreparedStatementMethodRule
tags: [phpstan.rules.rule]
arguments:
    classMethods:
        - 'My\Database::write'
```

The most generic example I can find is the following:
```php
$db->write('DELETE FROM my_table WHERE col1 = :col1 AND col2 = :col2', [
    'col1' => $value,
]);
```
This reports no error, even though the `col2` placeholder is missing. Furthermore, I can introduce an error into the query string by, e.g., using the wrong table name:
```php
$db->write('DELETE FROM table_does_not_exist WHERE col1 = :col1 AND col2 = :col2', [
    'col1' => $value,
]);
```
This will also report no error.

In either case, if I change `DELETE` to `SELECT`, all errors are reported:

```php
$db->write('SELECT FROM my_table WHERE col1 = :col1 AND col2 = :col2', [
    'col1' => $value,
]);
```
> Query expects placeholder :col2, but it is missing from values given.

```php
$db->write('SELECT FROM table_does_not_exist WHERE col1 = :col1 AND col2 = :col2', [
    'col1' => $value,
]);
```
> Query error: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'dbname.table_does_not_exist' doesn't exist (42S02).

Turning on debug mode (`$config->debugMode(true);`) doesn't provide any additional output.

If there is any additional information I can provide, please don't hesitate to let me know. Thanks for your time!

hemberger:
Remove the restriction that placeholder validation is performed only for SELECT queries. As a result, write queries will now validate their placeholders if write query analysis is enabled.

Previously, placeholder errors in a write query would result in all errors in the query being suppressed (even those unrelated to the placeholders).

Fixes #503.


tests/rules/SyntaxErrorInPreparedStatementMethodRuleTest.php
```diff
@@ -157,6 +161,10 @@ public function testSyntaxErrorInQueryRule(): void
                            ^ (42601).',
                     319,
                 ],
+                [
+                    'Query expects placeholder :freigabe1u1, but it is missing from values given.',
```
staabm, line 165: plase check the example in this file. 
```
$connection->preparedQuery('INSERT into ada(adaid, gesperrt, email,freigabe1u1) values(:adaid, :gesperrt, :email, :freigabe1u1)', ['adaid' => 1, 'gesperrt' => 0, 'email' => 'test@github.com', 0]);
```

I guess this error is not correct, since the last value of the array will be handled by PDO as the last value of the column list.

please also add new test-cases (at the very end of the file) which show that the rule works for writable queries as you expect it (and also cover cases which it should not error about)

hemberger: I think this query is invalid (because it names a parameter `:freigabe1u1` that it does not define), and the new warning is correct. In my testing, I get the following error if I actually try to execute this query:
> PDOException: SQLSTATE[HY093]: Invalid parameter number: parameter was not defined

The write queries that are already part of the test suite seem to cover all the cases I can think of. I'm happy to add more if you feel that some category has been omitted. Please let me know!

staabm: I see, thx for testing. we can merge after conflicts are resolved then.

staabm:
thank you

staabm:
your first feature is avaible within the new release:

https://twitter.com/markusstaab/status/1626234424059908096

thank you!

PR #511, opened 2023-02-16
Allow PHP 7.2

TomasVotruba:


staabm:
in case you don't want to fiddle with the php-cs-fixer, I would be fine changing to ECS in case its easier for you

TomasVotruba:
> I would be fine changing to ECS in case its easier for you

Me too :) but first things first. The `composer` command handles it well :+1:

TomasVotruba:
Not sure why locally the coding style passed and remotely not. Any ideas?

TomasVotruba:
Good, I re-run with different PHP version :) :+1:

TomasVotruba:
I need help with the SQL fail, as I have no idea where that does come from.

staabm:
I will push it over the finish line, thank you.

seems some kind of whitespace difference in error messages

TomasVotruba:
Thanks a lot :100: 

You're probably right. I assume the SQL is longer in `doctrine/dbal` 2.13

tests/rules/SyntaxErrorInQueryMethodSubclassedRuleTest.php
```diff
@@ -58,18 +58,18 @@ public function getExpectedErrors(): array
...
                     ],
                     [
                         <<<TEXT
-    Query error: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "freigabe1u1"
-    LINE 1: SELECT email adaid WHERE gesperrt freigabe1u1 FROM ada LIMIT...
-                                              ^ (42601).
+Query error: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "freigabe1u1"
+LINE 1: SELECT email adaid WHERE gesperrt freigabe1u1 FROM ada LIMIT...
+                                          ^ (42601).
 TEXT
```
TomasVotruba, line 73: I see :+1: 

I wonder if these would be better in txt files, as the spacing is crappy here and in PHP 7.3+ herenowdoc change can backfire again :)

staabm:
thanks you!

tests/rules/SyntaxErrorInQueryMethodSubclassedRuleTest.php
staabm: its easier to maintain in a single file instead of distributing it all over. its good enough for now :)

staabm:
pushed as a new release in https://github.com/staabm/phpstan-dba/releases/tag/0.2.57

williamdes:
awesome, thank you!

PR #512, opened 2023-02-16
Fix wrong deprecated annotation

mitelg:
Before the change PHPStorm suggest the misspelling in auto-completion which is a bit annoying :grin:


staabm:
ha, thank you

PR #523, opened 2023-02-17
Test PHP 7.2 in github action

staabm:

PR #528, opened 2023-02-17
Add ECS and apply standard sets

TomasVotruba:
- replace cs fixer with ecs
- update compose run
- [cs] apply PSR-12 and other sets


TomasVotruba:
As request yesterday :) 

I see the php-cs-fixer now requires PHP 7.3, which is a pitty. ECS solves this :+1: 

I've applied all typical comon sets, including PSR 12 and strict types. Let me know if there is something you really don't like.O

staabm:
Looks great. 🙏

TomasVotruba:
Awesome :+1:

staabm:
btw: I have also running a [green PHP 7.2 CI job](https://github.com/staabm/phpstan-dba/pull/523).

PR #529, opened 2023-02-19
Add DoctrineKeyValueStyleRule

### LINKED ISSUE #501: Support for insert/delete/update wrappers
As originally requested in https://github.com/staabm/phpstan-dba/issues/472#issuecomment-1368637186, I was wondering if it would be possible to support Doctrine-style wrappers for write operations.

For example, an `insert` method is converted into a query as follows:

```php
$conn->insert('user', ['username' => 'jwage']);
// INSERT INTO user (username) VALUES (?) (jwage)
```

This conversion is described explicitly in the help text for the following methods:

[Doctrine\DBAL\Connection::delete](https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/data-retrieval-and-manipulation.html#delete)
[Doctrine\DBAL\Connection::insert](https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/data-retrieval-and-manipulation.html#insert)
[Doctrine\DBAL\Connection::update](https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/data-retrieval-and-manipulation.html#update)

If supported, my expectation would be that PHPStan could identify if any of the table or column names did not exist.

In addition to the Doctrine methods, I also have custom wrappers that use a similar interface (i.e. table name as first argument, column name-value array for any additional arguments), but perform a query that Doctrine doesn't have a convenience wrapper for, e.g.

```php
$myconn->replace('user', ['username' => 'jwage']);
// REPLACE INTO user (username) VALUES (?) (jwage)
```
As such, I would find it very useful to have some generic way to support interfaces like these. I have no idea what would be easiest in terms of implementation, but I was imagining the following possibilities:
1. A `SyntaxError` class that could be specified in the PHPStan config, e.g.
    ```neon
    class: staabm\PHPStanDba\Rules\SyntaxErrorInQueryWrapperRule # hopefully w/ a better name!
    tags: [phpstan.rules.rule]
    arguments:
        classMethods:
            - 'My\Connection::replace'
    ```
2. Parameter types similar to `class-string` that could identify table/column names, e.g.
    ```php
    /**
     * @param phpstandba-table-string $table
     * @param array<phpstandba-column-string, mixed> $params
     */
    public function replace(string $table, array $params)
    ```

I'm happy to help in whatever way I can, though I should note that I'm a complete newbie when it comes to extending PHPStan, so would benefit greatly from some pointers if you want me to submit a PR. :)

Thank you for your time, and for all the work you've done on this amazing project. I'm super excited to finally have some confidence in my SQL queries!

hemberger:
This implements a new Rule for methods that build queries from simple
ingredients: a table name, and arrays of column name/value elements.
It checks that all table and column names are valid, and that the array
value types are (maybe) accepted by the column types.

For configuration, the class::method is specified as usual, with an
optional comma-delimited list of array agrument indices after the hash.
Specifying the indices is necessary because this rule covers 3 kinds of
simple builders:

1. No array arguments (for truncate-like queries)
2. One array argument (for insert/delete/replace-like queries)
3. Two array arguments (for update-like queries)

The most notable examples of this type of method are the Doctrine\DBAL
insert, delete, and update methods on their Connection object, which
have additional arguments that are not arrays of column names/values.

Closes #501.

---------------------------

A couple open questions about implementation for you:

* **Naming**. I chose "QueryAssembler" as the rule name to avoid confusing it with the Doctrine\DBAL QueryBuilder, which has a much more advanced API. An alternate name I was considering was "SimpleQueryBuilder", but I'm happy to change it to whatever you want.
* **Configuration**. The comma-delimited list of array argument indices is a new config syntax. This was the best I could come up with to support the existing Doctrine\DBAL methods. If you prefer to handle it a different way, just let me know.
* **Type checking**. Value types are validated against column types, which is a very useful check, but it is more strict than what is permissible by most DBALs (due to type coercion). It might warrant being a new option, which could enable type checking on parameters in prepared statements as well. Do you want me to remove type checking here and propose it in a separate PR?

Thanks again for your time, and I appreciate any feedback you can provide!


hemberger:
I don't understand why only the pdo-mysql replay test is failing. I might have expected that I need to update the cache files, but then I'd expect the mysqli replay test to fail as well. When I run the tests locally they make pretty significant changes to the cache files, so I'm hesitant to include them in the PR (most of it is adding a leading `\` to the object names for some reason...).

config/rules.neon
```diff
@@ -62,6 +62,18 @@ services:
...
                 - 'mysqli_execute_query#1'
 
+    -
+        class: staabm\PHPStanDba\Rules\SyntaxErrorInQueryAssemblerRule
+        tags: [phpstan.rules.rule]
+        arguments:
+            classMethods:
+                - 'staabm\PHPStanDba\Tests\Fixture\Connection::assembleNoArrays'
+                - 'staabm\PHPStanDba\Tests\Fixture\Connection::assembleOneArray#1'
+                - 'staabm\PHPStanDba\Tests\Fixture\Connection::assembleTwoArrays#1,2'
```
staabm, line 72: rule definition only meant for testing, should be declared in the `tests/rules/config/`-files.

in the `config/rules.neon` we should only have namespaces to real world definitions and functions.

src/Rules/SyntaxErrorInQueryAssemblerRule.php
```diff
@@ -0,0 +1,181 @@
...
+            }
+            if ($arrayArgPositions !== null) {
+                $arrayArgPositions = array_map('intval', explode(',', strval($arrayArgPositions)));
+            } else {
+                $arrayArgPositions = [];
+            }
+            $this->classMethods[] = [$className, $methodName, $arrayArgPositions];
+        }
+        $this->queryReflection = new QueryReflection();
+    }
```
staabm, line 59: I am not sure its worth to have this method this kind of configurable.

do we have identical api contracts on classes outside of doctrine/dbal?
I am wondering whether we should instead have 3 rules which map the identical needs of the 3 doctrine-dbal use-cases?

staabm (commented):
I like how the new SchemaReflection API makes these rules to implement that straight forward :)

config/rules.neon
hemberger: Hmm, I tried putting them in `tests/*/config/phpstan.neon.dist`, but then the phpunit tests all failed. Adding the services via `getAdditionalConfigFiles()` in the test classes gave me an error, e.g.
> Multiple services of type staabm\PHPStanDba\Rules\SyntaxErrorInQueryAssemblerRule found

Here's my attempt:

https://github.com/hemberger/phpstan-dba/commit/7b9c9d4e5b314e5a1f8d3b191b4fda0ce012b25b

Any idea what's wrong with that? I should note that all the existing test fixtures were already in `config/rules.neon`.

src/Rules/SyntaxErrorInQueryAssemblerRule.php
hemberger: If we were to implement them as separate rules with no arguments, I think they would be sufficiently similar that it would make sense to have an abstract base class with child classes that just define which arguments are column name/value arrays. Let me know if you'd like me to make that change, and if you have any suggestions for class names.

I don't know of any interfaces like this outside of Doctrine/DBAL. I have my own wrappers, but they follow the same format, e.g.
* truncate
* delete#1
* replace#1
* insert#1
* update#1,2

I can't imagine any other use cases. Maybe more automated SELECT queries, but probably better to use a fully-fledged query builder at that point.

staabm: I think it could make sense to have a single rule which can be used for any api which takes a table name as a string.
this could be a use case worth supporting for a separate rule (could be a separate PR though).

for the doctrine dba delete,insert,update api I think we should have a rule dedicated for that api, without configuration but hardcoded signatures

config/rules.neon
staabm: > I should note that all the existing test fixtures were already in config/rules.neon.

which ones do you mean exactly?

hemberger: The following are already in `config/rules.neon`.
```
 - 'staabm\PHPStanDba\Tests\Fixture\Connection::preparedQuery'
 - 'staabm\PHPStanDba\Tests\Fixture\PreparedStatement::__construct'
 - 'staabm\PHPStanDba\Tests\Fixture\BaseQueryClass::doQuery'
 - 'staabm\PHPStanDba\Tests\Fixture\BaseQueryClass::doQuery#0'
```

src/Rules/SyntaxErrorInQueryAssemblerRule.php
hemberger: Sure, I can make some changes. Just to confirm, instead of:
```neon
    -   
        class: staabm\PHPStanDba\Rules\SyntaxErrorInQueryAssemblerRule
        tags: [phpstan.rules.rule]
        arguments:
            classMethods:
                - 'Doctrine\DBAL\Connection::insert#1'
                - 'Doctrine\DBAL\Connection::delete#1'
                - 'Doctrine\DBAL\Connection::update#1,2'
                - 'My\Connection::truncate'
                - 'My\Connection::insert#1'
                - 'My\Connection::replace#1'
                - 'My\Connection::delete#1'
                - 'My\Connection::update#1,2'
```
You want:
```neon
    -   
        class: staabm\PHPStanDba\Rules\DoctrineErrorInInsertMethodRule
        tags: [phpstan.rules.rule]
    -   
        class: staabm\PHPStanDba\Rules\DoctrineErrorInDeleteMethodRule
        tags: [phpstan.rules.rule]
    -   
        class: staabm\PHPStanDba\Rules\DoctrineErrorInUpdateMethodRule
        tags: [phpstan.rules.rule]
    -   
        class: staabm\PHPStanDba\Rules\ErrorInQueryAssemblerMethodRule
        tags: [phpstan.rules.rule]
        arguments:
            classMethods:
                - 'My\Connection::truncate'
                - 'My\Connection::insert#1'
                - 'My\Connection::replace#1'
                - 'My\Connection::delete#1'
                - 'My\Connection::update#1,2'
```
with the arguments hard-coded for the Doctrine classes? Or do you want Doctrine users to be able to cover all three methods with a single class, e.g. `DoctrineErrorInQueryAssemblerMethodRule`?

staabm: I am fine with a single doctrine rule, or 3 separate ones.

If we have the doctrine rule we don't need the SyntaxErrorInQueryAssemblerRule anymore, right?

hemberger: At least for my use case, we'd still need the more generic rule, since I have my own wrappers (as enumerated by the `My\Connection` examples). But if you don't want that in the upstream repo, I can remove it.

staabm: Ok, got it now. Your wrappers class wrapps the doctrine api itself, or just uses the same style?

src/Rules/SyntaxErrorInQueryAssemblerRule.php
```diff
@@ -0,0 +1,181 @@
...
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\VerbosityLevel;
+use staabm\PHPStanDba\QueryReflection\QueryReflection;
+
+/**
+ * @implements Rule<CallLike>
+ *
+ * @see SyntaxErrorInQueryAssemblerRuleTest
+ */
+final class SyntaxErrorInQueryAssemblerRule implements Rule
```
staabm, line 28:
```suggestion
final class DoctrineKeyValueStyleRule implements Rule
```

src/Rules/SyntaxErrorInQueryAssemblerRule.php
staabm: See https://github.com/staabm/phpstan-dba/pull/529/files#r1112353477

config/rules.neon
staabm: We will clean this up separately

staabm:
thank you

src/Rules/SyntaxErrorInQueryAssemblerRule.php
hemberger: Sorry for the slow response. I can usually only get to this stuff on the weekends.

It looks like you were able to finish this up in a way that you were happy with. I've confirmed that it works for me. Thanks so much for your help and patience!

Issue #527, opened 2023-02-17
nullability and join types


staabm:
When e.g. outer joining a table a column which is defined as non-null can/will still contain null when not matched

We should add test coverage for all join cases

PR #531, opened 2023-02-20
Narrow types via join / join-condition

staabm:
closes https://github.com/staabm/phpstan-dba/issues/527

### DIFF
--- a/src/SchemaReflection/Join.php
+++ b/src/SchemaReflection/Join.php
@@ -14,34 +14,32 @@ final class Join
     public const TYPE_OUTER = 'outer';
 
     /**
-     * @var self::TYPE_*
+     * @var string
      */
-    private $type;
-
+    private $joinType;
     /**
      * @var Table
      */
     private $table;
 
     /**
-     * @param self::TYPE_* $type
+     * @param self::TYPE_* $joinType
      */
-    public function __construct(string $type, Table $table)
-    {
-        $this->type = $type;
+    public function __construct(string $joinType, Table $table) {
+        $this->joinType = $joinType;
         $this->table = $table;
-    }
-
-    /**
-     * @return self::TYPE_*
-     */
-    public function getType(): string
-    {
-        return $this->type;
     }
 
     public function getTable(): Table
     {
         return $this->table;
+    }
+
+    /**
+     * @return self::TYPE_*
+     */
+    public function getJoinType(): string
+    {
+        return $this->joinType;
     }
 }
--- a/src/SchemaReflection/Join.php
+++ b/src/SchemaReflection/Join.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace staabm\PHPStanDba\SchemaReflection;
+
+/**
+ * @api
+ */
+final class Join
+{
+    public const TYPE_INNER = 'inner';
+
+    public const TYPE_OUTER = 'outer';
+
+    /**
+     * @var self::TYPE_*
+     */
+    private $type;
+
+    /**
+     * @var Table
+     */
+    private $table;
+
+    /**
+     * @param self::TYPE_* $type
+     */
+    public function __construct(string $type, Table $table)
+    {
+        $this->type = $type;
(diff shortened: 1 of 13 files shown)


staabm:
@gharlan could you do me a favour and verify the expected types?

tests/default/data/sql-ast-narrowing.php
```diff
@@ -285,4 +285,34 @@ public function sum(PDO $pdo): void
...
+    public function joinNonNullable(PDO $pdo): void
+    {
+        $stmt = $pdo->query('SELECT adaid, akid from ada join ak on (adaid = eladaid)');
+        assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, akid: int<-2147483648, 2147483647>, 1: int<-2147483648, 2147483647>}>', $stmt);
+
+        $stmt = $pdo->query('SELECT adaid, akid from ada inner join ak on (adaid = eladaid)');
+        assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, akid: int<-2147483648, 2147483647>, 1: int<-2147483648, 2147483647>}>', $stmt);
+
+        $stmt = $pdo->query('SELECT adaid, akid from ada left join ak on (adaid = eladaid)');
+        assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, akid: int<-2147483648, 2147483647>|null, 1: int<-2147483648, 2147483647>|null}>', $stmt);
```
staabm, line 313: rows not matched in the main result, will be filled up with null, therefore even a non-nullable column can get null on outer joins

tests/default/data/sql-ast-narrowing.php
```diff
@@ -285,4 +285,34 @@ public function sum(PDO $pdo): void
         $stmt = $pdo->query('SELECT sum(c_double) as sum from typemix');
         assertType('PDOStatement<array{sum: float, 0: float}>', $stmt);
     }
+
+    public function joinNullable(PDO $pdo): void
+    {
+        $stmt = $pdo->query('SELECT adaid, eladaid from ada join ak on (adaid = eladaid)');
+        assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, eladaid: int<-2147483648, 2147483647>, 1: int<-2147483648, 2147483647>}>', $stmt);
+
+        $stmt = $pdo->query('SELECT adaid, eladaid from ada inner join ak on (adaid = eladaid)');
```
staabm, line 294: on inner join, nullable columns get non nullable

gharlan (approved):
I think these expectations are correct 👍

src/SchemaReflection/Join.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace staabm\PHPStanDba\SchemaReflection;
+
+/**
+ * @api
+ */
+final class Join
```
staabm, line 10: not sure it makes sense to model joins here. maybe we should instead use `SqlFtw\Sql\Dml\TableReference\Join` instead, as we need the low level information

src/SqlAst/QueryScope.php
```diff
@@ -106,10 +107,15 @@ public function getType($expression): Type
...
+            foreach ($this->joinedTables as $join) {
+                $joinedTable = $join->getTable();
+
                 foreach ($joinedTable->getColumns() as $column) {
                     if ($column->getName() === $expression->getName()) {
-                        return $column->getType();
+                        if ($join->getType() === Join::TYPE_OUTER) {
+                            return TypeCombinator::addNull($column->getType());
+                        }
+                        return TypeCombinator::removeNull($column->getType());
```
staabm, line 118: I think we should only add/remove NULLs for columns which are part of the join condition, but not regular columns selected from joined tables

staabm: before doing so, we may wait a bit for sqlftw stabilization as we would need a deep dive into the AST

grafik

PR #557, opened 2023-03-01
prevent excessive sql queries in DoctrineKeyValueStyleRule

staabm:


staabm:
//cc @hemberger

hemberger:
This works great, and the plan for next major release sounds good too. Thank you!

Issue #560, opened 2023-03-03
Argument expects a literal string, got literal-string


noemi-salaun:
When calling `Connection::insert()` with the table name as a `literal-string`, I got this error message

> Argument #0 expects a literal string, got literal-string

staabm:
thanks for reporting. could you please provide a concrete code sample?

noemi-salaun:
I have some code that clone a user with all its data.

1. So I got an hardcoded list of known tables.
2. Foreach on the list.
3. SELECT * FROM each table
4. Change some of the data (replace the userId used as a foreign key, ...)
5. INSERT the new data in the same table

For the insert, I use a function that add backquote around column and table name.

```php
/**
 * @param literal-string       $tableExpression
 * @param array<string, mixed> $data
 */
private function insert(string $tableExpression, array $data): void
{
    $backquotedTableExpression = "`$tableExpression`";

    $backquotedData = [];
    foreach ($data as $key => $value) {
        $backquotedData["`$key`"] = $value;
    }

    $this->dbal->insert($backquotedTableExpression, $backquotedData);
}
```

So my real error message is

> Argument#0 expects a literal string, got literal-string&non-falsy-string

But the error stay the same if I remove the backquote and pass `$tableExpression` directly to dbal insert

> Argument#0 expects a literal string, got literal-string

hemberger:
What is being checked in the `DoctrineKeyValueStyleRule` is that the keys of the array passed to `Doctrine\DBAL\Connection::insert` correspond to columns in the database. In your `insert` function, PHPStan doesn't know the literal values of the keys of `$data`, and therefore it cannot check that they are columns in the database.

If you are wrapping calls to `Doctrine\DBAL\Connection` in your own class, e.g. `My\Connection`, then you should probably be declaring its function in your PHPStan configuration instead of the Doctrine ones. For example:

```neon
services:
    -
        class: staabm\PHPStanDba\Rules\DoctrineKeyValueStyleRule
        tags: [phpstan.rules.rule]
        arguments:
            classMethods:
                - 'My\Connection::insert#1'
```

(Note: technically this rule checks for a _constant_ string, so maybe the error message should be revised to replace "literal" -> "constant"?)

staabm:
> (Note: technically this rule checks for a _constant_ string, so maybe the error message should be revised to replace "literal" -> "constant"?)

ohh I did not see that. we should not talk about `literal` strings when we expect `constant`  strings and vice versa.
[...]

PR #564, opened 2023-03-11
DoctrineKeyValueStyleRule: improve table error message

hemberger:
Table names must be constant strings, not literal strings.

As per https://github.com/staabm/phpstan-dba/issues/560#issuecomment-1464843621.

### DIFF
--- a/src/Rules/DoctrineKeyValueStyleRule.php
+++ b/src/Rules/DoctrineKeyValueStyleRule.php
@@ -111,7 +111,7 @@ final class DoctrineKeyValueStyleRule implements Rule
         $tableType = $scope->getType($tableExpr);
         if (! $tableType instanceof ConstantStringType) {
             return [
-                RuleErrorBuilder::message('Argument #0 expects a literal string, got ' . $tableType->describe(VerbosityLevel::precise()))->line($callLike->getLine())->build(),
+                RuleErrorBuilder::message('Argument #0 expects a constant string, got ' . $tableType->describe(VerbosityLevel::precise()))->line($callLike->getLine())->build(),
             ];
         }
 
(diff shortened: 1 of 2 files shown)


staabm:
thanks

src/Rules/DoctrineKeyValueStyleRule.php
```diff
@@ -111,7 +111,7 @@ public function processNode(Node $callLike, Scope $scope): array
         $tableType = $scope->getType($tableExpr);
         if (! $tableType instanceof ConstantStringType) {
```
staabm, line 112: reading this again 2 things come to mind
1) I think we should work with `$tableType->getConstantStrings()` which would allow things like
```
    public function errorIfTableDoesNotExist(Connection $conn)
    {
        $table = 'ada';
        if (rand(0, 1) === 1) {
           $table = 'ak';
        }
        $conn->assembleNoArrays($table);
    }
```

2) I think we should onyl return a error when debug-mode is enabled. without debug mode we should just accept a not analyzable situation and return without a error

hemberger: Good idea, thanks for the suggestion! I've started working on removing all uses of `instanceof` in this class, but will wait to finalize it until after #563 is resolved (to avoid massive merge conflicts).

hemberger: > 2. I think we should only return a error when debug-mode is enabled. without debug mode we should just accept a not analyzable situation and return without a error

@staabm I'm working on this request now. Would you also like to consider mixed value types as "not analyzable" so that errors such as the following only show up in debug mode?

```
Column 'foo.id' expects value type int, got type mixed
```
Or should I leave these as always enabled? Thanks!

PR #569, opened 2023-03-15
support for cross joins. don't error on joins of subqueries

jakubvojacek:
- support for multiple joins, up to now it was OK just for single join, this adds support for multiple
 - skips AST for 
    - cross joins
    - queries with subqueries
    - when condition is null (previous it was throwing runtime exception, now it just screams in debug)

The changes works fine for my codebase

### DIFF
--- a/src/SqlAst/ParserInference.php
+++ b/src/SqlAst/ParserInference.php
@@ -17,8 +17,6 @@ use SqlFtw\Sql\Dml\Query\SelectCommand;
 use SqlFtw\Sql\Dml\Query\SelectExpression;
 use SqlFtw\Sql\Dml\TableReference\InnerJoin;
 use SqlFtw\Sql\Dml\TableReference\Join;
-use SqlFtw\Sql\Dml\TableReference\JoinSide;
-use SqlFtw\Sql\Dml\TableReference\OuterJoin;
 use SqlFtw\Sql\Dml\TableReference\TableReferenceSubquery;
 use SqlFtw\Sql\Dml\TableReference\TableReferenceTable;
 use SqlFtw\Sql\Expression\Identifier;
--- a/src/SqlAst/ParserInference.php
+++ b/src/SqlAst/ParserInference.php
@@ -24,7 +24,7 @@ use SqlFtw\Sql\SqlSerializable;
 use staabm\PHPStanDba\QueryReflection\QueryReflection;
 use staabm\PHPStanDba\SchemaReflection\Join as SchemaJoin;
 use staabm\PHPStanDba\SchemaReflection\SchemaReflection;
-use staabm\PHPStanDba\UnresolvableQueryMixedTypeException;
+use staabm\PHPStanDba\UnresolvableAstInQueryException;
 
 final class ParserInference
 {
@@ -67,7 +67,7 @@ final class ParserInference
                     while (1) {
                         if ($from->getCondition() === null) {
                             if (QueryReflection::getRuntimeConfiguration()->isDebugEnabled()) {
-                                throw new UnresolvableQueryMixedTypeException('Cannot narrow down types null join conditions: ' . $queryString);
+                                throw new UnresolvableAstInQueryException('Cannot narrow down types null join conditions: ' . $queryString);
                             }
 
                             return $resultType;
@@ -75,7 +75,7 @@ final class ParserInference
 
                         if ($from->getRight() instanceof TableReferenceSubquery || $from->getLeft() instanceof TableReferenceSubquery) {
                             if (QueryReflection::getRuntimeConfiguration()->isDebugEnabled()) {
-                                throw new UnresolvableQueryMixedTypeException('Cannot narrow down types for SQLs with subqueries: ' . $queryString);
+                                throw new UnresolvableAstInQueryException('Cannot narrow down types for SQLs with subqueries: ' . $queryString);
                             }
 
                             return $resultType;
@@ -83,7 +83,7 @@ final class ParserInference
 
                         if ($from instanceof InnerJoin && $from->isCrossJoin()) {
                             if (QueryReflection::getRuntimeConfiguration()->isDebugEnabled()) {
-                                throw new UnresolvableQueryMixedTypeException('Cannot narrow down types for cross joins: ' . $queryString);
+                                throw new UnresolvableAstInQueryException('Cannot narrow down types for cross joins: ' . $queryString);
                             }
 
                             return $resultType;
--- a/src/SqlAst/ParserInference.php
+++ b/src/SqlAst/ParserInference.php
@@ -17,11 +17,16 @@ use SqlFtw\Sql\Dml\Query\SelectCommand;
 use SqlFtw\Sql\Dml\Query\SelectExpression;
 use SqlFtw\Sql\Dml\TableReference\InnerJoin;
 use SqlFtw\Sql\Dml\TableReference\Join;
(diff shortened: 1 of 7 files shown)


jakubvojacek:
> changes look good to me, but require test-cases which fail without the actual code changes

yea, I will try to add some

tests/default/data/sql-ast-narrowing.php
```diff
@@ -365,4 +365,18 @@ public function joinSelectOutsideCondition(PDO $pdo): void
         assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, eadavk: numeric-string|null, 1: numeric-string|null}>', $stmt);
     }
 
+    public function multipleJoins(PDO $pdo): void
+    {
+        $stmt = $pdo->query('SELECT adaid from ada join ak on adaid = eladaid inner join typemix on adaid = c_int');
```
staabm, line 370: Please also select columns from within the join condition, make sure these are narrowed properly and make sure folumns from joined tables not part of the condition are also narrowed properly.

jakubvojacek: I tried, please check, is that you you had in mind?

src/SqlAst/ParserInference.php
```diff
@@ -61,27 +64,57 @@ public function narrowResultType(string $queryString, ConstantArrayType $resultT
...
+                            return $resultType;
+                        }
+
+                        if ($from instanceof InnerJoin && $from->isCrossJoin()) {
+                            if (QueryReflection::getRuntimeConfiguration()->isDebugEnabled()) {
+                                throw new UnresolvableQueryMixedTypeException('Cannot narrow down types for cross joins: ' . $queryString);
+                            }
+
+                            return $resultType;
+                        }
```
staabm, line 90: why do we need these early returns?

tests/default/data/sql-ast-narrowing.php
```diff
@@ -365,4 +365,21 @@ public function joinSelectOutsideCondition(PDO $pdo): void
         assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, eadavk: numeric-string|null, 1: numeric-string|null}>', $stmt);
     }
 
+    public function multipleJoins(PDO $pdo): void
+    {
+        $stmt = $pdo->query('SELECT adaid, eladaid, c_int, c_char5 from ada inner join ak on adaid = eladaid inner join typemix on adaid = c_int');
+        assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, eladaid: int<-32768, 32767>, 1: int<-32768, 32767>, c_int: int<-32768, 32767>, 2: int<-32768, 32767>, c_char5: string, 3: string}>', $stmt);
+
+        $stmt = $pdo->query('SELECT adaid, eladaid, c_int, c_char5 from ada left join ak on adaid = eladaid left join typemix on adaid = c_int');
+        assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, eladaid: int<-32768, 32767>|null, 1: int<-32768, 32767>|null, c_int: int<-32768, 32767>|null, 2: int<-32768, 32767>|null, c_char5: string|null, 3: string|null}>', $stmt);
```
staabm, line 374: i think the test expectations look correct. 

did they make sense to you, too? or did you just copy the current results from the test-suite and have questions about it?

src/SqlAst/ParserInference.php
jakubvojacek: it cannot collect proper data about joins and the columns hence it gives up on narrowing down the type

tests/default/data/sql-ast-narrowing.php
jakubvojacek: It makes sense to me as well, was just making sure that this is what you meant.

src/SqlAst/ParserInference.php
staabm: but would it make a difference for the added test-cases if we just drop it here. wouldn't it give up after the loop then?

tests/default/data/sql-ast-narrowing.php
```diff
@@ -365,4 +365,21 @@ public function joinSelectOutsideCondition(PDO $pdo): void
...
+    }
+
+    public function ignoredAstQueries(PDO $pdo): void
+    {
+        $pdo->query('SELECT adaid from ada inner join (select akid from ak) on akid = adaid');
```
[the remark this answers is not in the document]
jakubvojacek: because it has a different output for `pdo-mysql` and `mysqli` reflectors (but that is unrelated to AST analysis since we give up on that due to the subquery)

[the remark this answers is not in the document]
staabm: without an assertion this line does not have value. what is the difference between the reflectors regarding this statement?

src/SqlAst/ParserInference.php
jakubvojacek: without the early return, it wouldnt work, the tests would fail

```
1) Error
The data provider specified for staabm\PHPStanDba\Tests\DbaInferenceTest::testFileAsserts is invalid.
Error: Call to undefined method SqlFtw\Sql\Dml\TableReference\TableReferenceSubquery::getTable()
/Users/jakubvojacek/Sites/phpstan-dba/src/SqlAst/ParserInference.php:96
/Users/jakubvojacek/Sites/phpstan-dba/src/QueryReflection/QueryReflection.php:135
/Users/jakubvojacek/Sites/phpstan-dba/src/PdoReflection/PdoStatementReflection.php:88
/Users/jakubvojacek/Sites/phpstan-dba/src/Extensions/PdoQueryDynamicReturnTypeExtension.php:106
```

jakubvojacek: but that will trigger phpstan internal error and the analysis will fail and phpstan cache wont be saved,right?I originally had the exception there but it was so annoying when running on my codebase that I decided to hide it behind debug flag.

staabm: I think it should work when we create a new subclass of `UnresolvableQueryException`

similar to how we did with https://github.com/staabm/phpstan-dba/blob/main/src/UnresolvableQueryMixedTypeException.php which seems to work for you atm

tests/default/data/sql-ast-narrowing.php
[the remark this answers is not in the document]
jakubvojacek: my bad, must have been doing something wrong back then when i was testing. I just pushed a new assertion, hopefully CI will be green

src/SqlAst/ParserInference.php
jakubvojacek: currently

staabm: e.g. `UnresolvableAstInQueryException`

jakubvojacek: so you want me to rename `UnresolvableQueryMixedTypeException` to `UnresolvableAstInQueryException` on those three places that it thrown in ParserInference right?

I checked where UnresolvableQueryMixedTypeException was used and its only on one place and its after the `QueryReflection::getRuntimeConfiguration()->isDebugEnabled()` condition as well

staabm: nope, I want you to create a new exception class and use it here in this few places of this PR, but leave `UnresolvableQueryMixedTypeException`  as is across the codebase

staabm (approved):
thank you

Issue #566, opened 2023-03-14
Query AST analysis?


jakubvojacek:
Hello @staabm 

how far advanced do you think the analysis is, should we start reporting issues? I get errors thrown on some more non-standard queries, such as `CROSS join` or queries with subqueries. 

But I also tested with simple `select count(*) from xx` and I didnt get a `RangeIntegerType`. I tried digging deeper and found that it never got past this condition https://github.com/staabm/phpstan-dba/blob/main/src/SqlAst/ParserInference.php#L102 (`$valueType` was `min: 0, max: null` but it was not merged to `$resultType` due to `hasOffsetValueType === false`)

I am not sure I understand all the parts of the code to start sending PRs though.

Thanks

staabm:
hey,

thx for asking. for SQL AST to work you need to enable it in your `phpstan-dba-bootstrap` via [`RuntimeConfiguration`](https://github.com/staabm/phpstan-dba/blob/6b8f11251db4f023bfcb2449f594426605d70f81/src/QueryReflection/RuntimeConfiguration.php#L137-L147)

after thats done and you are using mysql it should work for `select count(*) from xx`.
I would love someone doing more testing with it.

> such as `CROSS join` or queries with subqueries.

I have open todos regarding joins, which I have expressed here https://github.com/staabm/phpstan-dba/issues/527
if you have concrete examples and concrete expectations - at best already in form of a unit test - that would be really awesome.

jakubvojacek:
>  you need to enable it in your phpstan-dba-bootstrap 

yeah, I did that

> please provide a reproducing case.

will try, perhaps I am still missunderstanding some parts of it and coming up with repro case will help me understand.

Thanks :+1:

PR #570, opened 2023-03-15
Failing AST test with non-named column

jakubvojacek:
Hello @staabm 

this is related to https://github.com/staabm/phpstan-dba/issues/566#issuecomment-1468512616, this is the reproducing case. When `count(*)` (or other functions) is used without alias, it fails to narrow down the type for integer indexed key in the resulting PDOStatement. 

I tried playing around with ParserInference but couldn't find a easy way to fix this yet


staabm:
to implement this I think we need https://github.com/SQLFTW/sqlftw/commit/da4e0354e6d48dffb94a04a316b1b08b6a2ca38e which will be part of the next sqlftw release

jakubvojacek:
perfect, cant wait for this

staabm:
Today a new sqlftw was released. You can give it a try and update the lob with this PR and add a fix for you problem if you like

jakubvojacek:
I will give it a go over the weekend! Thanks

jakubvojacek:
The fix for the correct type inference was easy, but tests are failing @staabm 

We got some problems with postgres tests, since AST does not work there, it fails (https://github.com/staabm/phpstan-dba/actions/runs/4454918527/jobs/7824413966?pr=570)

For the same reason, it also fails for php 7.3 and lower

Any idea how to overcome this, please?

staabm:
I think we should move 

https://github.com/staabm/phpstan-dba/blob/a1bb0c6e77f514528471e87d785084c21868b923/tests/default/DbaInferenceTest.php#L66-L68

into a separate test-class and enable the runtime-config for SQL AST only for this new test-class

jakubvojacek:
Is this what you had in mind please?

tests/stringify/data/ast-narrowed-stringify.php
```diff
@@ -13,6 +13,6 @@ public function count(PDO $pdo): void
         assertType('PDOStatement<array{myemail: numeric-string|null, 0: numeric-string|null}>', $stmt);
 
         $stmt = $pdo->query('SELECT count(email) as myemail, count(email) from ada');
-        assertType('PDOStatement<array{myemail: numeric-string|null, 0: numeric-string|null, count(email): numeric-string, 1: numeric-string|null}>', $stmt);
+        assertType('PDOStatement<array{myemail: numeric-string|null, 0: numeric-string|null, count(email): numeric-string|null, 1: numeric-string|null}>', $stmt);
```
staabm, line 16: I think this test no longer runs with SQL AST enabled

staabm (changes requested):
I don't think we need a whole new `defaultAst` suite, but just a new `SqlAstInferenceTest.php` within the `default` test-suite, which enables sql-ast like on setup/teardown, like we did for paramter validation in https://github.com/staabm/phpstan-dba/commit/bb2b8457f8a892f9ee59bc0fd1bdc285564c2292#diff-428d052928087e660593dfd032cc0445056ba9fe2780ddb2292983fefe04ac87R18-R26

PR #572, opened 2023-03-16
MysqlTypeMapper: add support for FLOAT types

hemberger:
Note that, unlike DOUBLE, the FLOAT type can be either single-precision or double-precision, depending on the precision specification. See: https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html.

Since PHP floats are double precision, you could technically run into issues when passing a PHP float into a MySQL FLOAT. However, since we only have a single `FloatType` in PHPStan, I don't think we have the ability to account for this.


staabm:
thank you

staabm:
the mysqli mapper also utilizes the mysql mapper. please add another test-case for the mysqli case
https://github.com/staabm/phpstan-dba/blob/90473e78a3ae303a3e5832474c8c0ede1b4ae1f4/src/TypeMapping/MysqliTypeMapper.php#L51

do you have experience with pgsql? do we need a similar change in `PgsqlTypeMapper`?

staabm:
thanks for looking into it. I have no personal experience either :).

@p4veI since you implemened pgsql support.. do you think we need something similar?

PR #573, opened 2023-03-16
Add mysqli test of MysqlTypeMapper

hemberger:
This mirrors the test in pdo-mysql.php, but with the mysqli driver instead of PDO. It tests all the types in the `typemix` table.

Requested in https://github.com/staabm/phpstan-dba/pull/572#issuecomment-1471467377.


hemberger:
I'll be honest, I don't really understand why a postgres test is failing here, since the only change is an added mysqli test. :thinking:

staabm:
I think we missed to update the https://github.com/staabm/phpstan-dba/blob/main/tests/pgsql-schema.sql

hemberger:
Yeah, the `typemix` table is really different between the two schemas. I spent quite a while trying to make them equivalent, but there are some differences between pgsql and mysql that can't be avoided, for example, pgsql doesn't have an `UNSIGNED` modifier.

My takeaway is that this particular test case shouldn't be run in the postgres tests. Do you want me to do this, or would you rather I leave it as-is so someone else can investigate?

staabm:
I think we should run the test but have different expectations depending on the used db backend

hemberger:
Thanks for the recommendation. This should be ready for review again.

tests/default/data/typemix-mysql.php
```diff
@@ -0,0 +1,24 @@
...
+
+    public function typemixMysqli(mysqli $mysqli)
+    {
+        $result = mysqli_query($mysqli, 'SELECT * FROM typemix');
+        assertType('mysqli_result<' . self::MYSQL_DATATYPES . '>', $result);
+    }
+
+    public function typemixPdoPgsql(PDO $pdo)
```
staabm, line 19: I wonder why the `typemix-mysql` test contains `typemixPdoPgsql` and the `typemix-pgsql` contains a `typemixPdoMysql`?

hemberger: Oops, good catch. I just got the function names backwards :)

Feel free to swap the names if you want to merge tonight. Otherwise, I'll do it tomorrow.

tests/default/data/typemix-pgsql.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    private const PGSQL_DATATYPES = 'array{pid: int<1, 2147483647>, c_varchar5: string, c_varchar25: string|null, c_varchar255: string, c_date: string|null, c_time: string|null, c_datetime: string|null, c_timestamp: string|null, c_text: string|null, c_enum: mixed, c_bit255: int, c_bit25: int|null, c_bit: int|null, c_int: int<-2147483648, 2147483647>, c_smallint: int<-32768, 32767>, c_bigint: int, c_float: float, c_boolean: bool, c_json: string, c_json_nullable: string|null, c_jsonb: string, c_jsonb_nullable: string|null}';
+
+    public function typemixMysqli(mysqli $mysqli)
+    {
+        $result = mysqli_query($mysqli, 'SELECT * FROM typemix');
+        assertType('mysqli_result<' . self::PGSQL_DATATYPES . '>', $result);
+    }
+
+    public function typemixPdoMysql(PDO $pdo)
```
staabm, line 19:
```suggestion
    public function typemixPdoPgsql(PDO $pdo)
```

staabm:
thank you

PR #575, opened 2023-03-16
DoctrineKeyValueStyleRule: properly handle unions

hemberger:
If any of the arguments (table name or parameter values) to the DoctrineKeyValueStyleRule methods are a union type, each subtype will now be checked independently.

Previously, union types would cause an error, because of the improper use of `instanceof *Type` checks. This is now fixed by using methods to get the constants, e.g. `getConstantStrings()`, as recommended in https://github.com/staabm/phpstan-dba/pull/564#discussion_r1133053061.

### DIFF
--- a/src/Rules/DoctrineKeyValueStyleRule.php
+++ b/src/Rules/DoctrineKeyValueStyleRule.php
@@ -14,8 +14,6 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleError;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\ShouldNotHappenException;
-use PHPStan\Type\Constant\ConstantArrayType;
-use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\IntegerRangeType;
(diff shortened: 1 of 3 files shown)


src/Rules/DoctrineKeyValueStyleRule.php
```diff
@@ -126,74 +125,81 @@ public function processNode(Node $callLike, Scope $scope): array
...
-            // If the argument doesn't exist, just skip it since we don't want
-            // to error in case it has a default value
-            if (! \array_key_exists($arrayArgPosition, $args)) {
-                continue;
-            }
-
-            $argType = $scope->getType($args[$arrayArgPosition]->value);
-            if (! $argType instanceof ConstantArrayType) {
-                $errors[] = 'Argument #' . $arrayArgPosition . ' is not a constant array, got ' . $argType->describe(VerbosityLevel::precise());
+        foreach ($tableNames as $tableName) {
```
staabm, line 129: I think this logic could benefit from extracting some parts in private methods.

atm its quite un-readable

staabm:
thank you

PR #580, opened 2023-03-18
AST named column (use raw expression from sqlwtf)

jakubvojacek:
https://github.com/staabm/phpstan-dba/pull/570 got too messy after rebasing etc, decided to make it again from scratch

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
 		"vlucas/phpdotenv": "^5.4"
 	},
 	"conflict": {
-		"sqlftw/sqlftw": "<0.1.6 || >0.1.7"
+		"sqlftw/sqlftw": "<0.1.10"
 	},
 	"autoload": {
 		"psr-4": {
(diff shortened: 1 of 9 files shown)


jakubvojacek:
I dont think this 

```php
protected function setUp(): void
    {
        QueryReflection::getRuntimeConfiguration()->utilizeSqlAst(true);
    }

    protected function tearDown(): void
    {
        QueryReflection::getRuntimeConfiguration()->utilizeSqlAst(false);
    }

```

works properly, the tests are failing as if the AST wasnt enabled at all (which I double checked in QueryReflector by dumping the runtime config)

```php
if (
                self::getRuntimeConfiguration()->isUtilizingSqlAst()
            ) {
```

I am not very much familiar wiht phpunit, I usually use nette/tester or codeception

staabm:
I think the reason that it worked for `DoctrineKeyValueStyleRuleStrictTest extends RuleTestCase`
but not in this PR is the different base class `DbaAstInferenceTest extends TypeInferenceTestCase`.
unfortunately we cannot use a different base class here.

it seems your previous approach using a new test-suite was the right one. I have forgotten about that. sorry.

PR #581, opened 2023-03-18
DoctrineKeyValueStyeRule: allow int values in float columns

hemberger:
Using `$databaseType->isSuperTypeOf($valueType)` to check if the input value type matches the database column type was too strict. It was incorrectly raising errors for various legal scenarios, e.g.:

* Passing an integer to a float column.
* Passing a benevolent (float|int) to an int column.

The `accepts` method correctly handles these cases. Add extra logic for `MixedType`, since mixed types are always accepted.

NOTE: The `MixedType` case may be demoted to a debug-mode-only error, depending on the answer to https://github.com/staabm/phpstan-dba/pull/564#discussion_r1141097432.

### DIFF
--- a/src/Rules/DoctrineKeyValueStyleRule.php
+++ b/src/Rules/DoctrineKeyValueStyleRule.php
@@ -171,11 +171,11 @@ final class DoctrineKeyValueStyleRule implements Rule
                                 continue;
                             }
 
-                            $acceptedType = $this->getColumnAcceptsType($argColumn->getType());
+                            $acceptingType = $this->getColumnAcceptingType($argColumn->getType());
                             $valueType = $argArray->getValueTypes()[$keyIndex];
 
-                            if (! $acceptedType->isSuperTypeOf($valueType)->yes()) {
-                                $errors[] = 'Column "' . $table->getName() . '.' . $argColumnName . '" expects value type ' . $acceptedType->describe(VerbosityLevel::precise()) . ', got type ' . $valueType->describe(VerbosityLevel::precise());
+                            if (! $acceptingType->isSuperTypeOf($valueType)->yes()) {
+                                $errors[] = 'Column "' . $table->getName() . '.' . $argColumnName . '" expects value type ' . $acceptingType->describe(VerbosityLevel::precise()) . ', got type ' . $valueType->describe(VerbosityLevel::precise());
                             }
                         }
                     }
@@ -194,7 +194,7 @@ final class DoctrineKeyValueStyleRule implements Rule
      * Converts the column type into the most general type that the column
      * will accept.
      */
-    private function getColumnAcceptsType(Type $columnType): Type
+    private function getColumnAcceptingType(Type $columnType): Type
     {
         $checkIntegerRanges = QueryReflection::getRuntimeConfiguration()->isParameterTypeValidationStrict();
         if (false === $checkIntegerRanges) {
--- a/src/Rules/DoctrineKeyValueStyleRule.php
+++ b/src/Rules/DoctrineKeyValueStyleRule.php
@@ -171,7 +171,7 @@ final class DoctrineKeyValueStyleRule implements Rule
                                 continue;
                             }
 
-                            $acceptedType = $this->columnAcceptsType($argColumn->getType());
+                            $acceptedType = $this->getColumnAcceptsType($argColumn->getType());
                             $valueType = $argArray->getValueTypes()[$keyIndex];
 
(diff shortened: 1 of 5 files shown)


src/Rules/DoctrineKeyValueStyleRule.php
```diff
@@ -194,7 +195,7 @@ public function processNode(Node $callLike, Scope $scope): array
                                 }
                             }
 
-                            if (! $argColumnType->isSuperTypeOf($valueType)->yes()) {
+                            if (! $argColumnType->accepts($valueType, true)->yes() || $valueType instanceof MixedType) {
```
staabm, line 198: `accepts´ does more than you actually want.

we should stay with `isSuperTypeOf`, but we could e.g. add things like
```
if ($argColumnType->isFloat()->yes()) {
  $numberType = new UnionType([new IntegerType(), new FloatType()]);
  return !$numberType->isSuperTypeOf($valueType)->yes();
}
```

to super float|int for float.

staabm:
> Passing a benevolent (float|int) to an int column.

do you say a regular `float|int` works but a benevolent `(float|int)` does not?
I kind of can't believe it :)

src/Rules/DoctrineKeyValueStyleRule.php
hemberger: What is `accepts` for if not this kind of logic? There's no documentation that I can find for the method, so if you can elaborate on its intended use, that would help me understand. Thanks!

hemberger:
> do you say a regular `float|int` works but a benevolent `(float|int)` does not? I kind of can't believe it :)

Given an `int` column, `float|int` is an error (as it should be), but currently `(float|int)` is also an error, and I was thinking that perhaps it shouldn't be given this documentation about them:

> Benevolent unions aren’t checked strictly even at the highest level

But if you say otherwise, I'll take your word for it! :)

src/Rules/DoctrineKeyValueStyleRule.php
staabm: accepts is meant to check whether a type at hand would accept another one, based on a lot of different phpstan configs (e.g. analysis level, feature flags, ..).

what we need is a way to determine whether the doctrine dbal api accepts a certain type and whether we can successfully store it in the db

staabm:
> Benevolent unions aren’t checked strictly even at the highest level

i guess you took that sentence from the phpstan docs somewhere? If so: it describes how native phpstan rules behave according to analysis level.

atm the phpstan-dba rules do not at all depend on phpstan rule levels

src/Rules/DoctrineKeyValueStyleRule.php
hemberger: Okay, thanks for the clarification. It sounds like the best course of action for now is to go with your suggested change. I'll work on that later today. Thanks again!

src/Rules/DoctrineKeyValueStyleRule.php
```diff
@@ -209,4 +189,33 @@ public function processNode(Node $callLike, Scope $scope): array
         }
         return $ruleErrors;
     }
+
+    /**
+     * Converts the column type into the most general type that the column
+     * will accept.
+     */
+    private function columnAcceptsType(Type $columnType): Type
```
staabm, line 197:
```suggestion
    private function getColumnAcceptsType(Type $columnType): Type
```

tests/rules/data/doctrine-key-value-style.php
```diff
@@ -119,4 +119,25 @@ public function errorInParamArrayUnion(Connection $conn, array $params)
...
+        $conn->assembleOneArray('typemix', ['c_int' => $value]);
+    }
+
+    /**
+     * @param array-key $value A benevolent union type (int|string)
+     */
+    public function errorWithBenevolentUnionValue(Connection $conn, $value)
+    {
+        $conn->assembleOneArray('typemix', ['c_int' => $value]);
+    }
```
staabm, line 137: I am not sure about this expectation. I would expect a error for a non-benevolent union, but I think we should be ok with a benevolent union, as long as it contains the type we expect

hemberger: I'm happy to change the behavior for benevolent unions as well. Would you be okay with me submitting it as a separate PR?

staabm:
thank you

tests/rules/data/doctrine-key-value-style.php
staabm: fine

PR #583, opened 2023-03-20
Fix mysql type mapper

jakubvojacek:
as discussed here https://github.com/staabm/phpstan-dba/pull/582#issuecomment-1475752161, the mapping from mysql native type to phpstan type was not working properly. 

This PR attempts to fix that


jakubvojacek:
not sure about the phpstan failures in tests @staabm, i guess they're related to php version?

hemberger:
I agree with this change, because MySQL explicitly states:
> The display width does not constrain the range of values that can be stored in the column. Nor does it prevent values wider than the column display width from being displayed correctly.

(source: https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html)

staabm:
> not sure about the phpstan failures in tests

I think this error was ignored before here

https://github.com/staabm/phpstan-dba/blob/154022382a7f809c8ddfc51e5241d8b62010d1f8/phpstan.neon.dist#L28

please adjust the ignore rule

jakubvojacek:
finally green :)

PR #584, opened 2023-03-20
SQL AST: COUNT cannot return null

hemberger:
Remove nullability from the PositiveIntReturnTypeExtension for the `COUNT` function case.

> If there are no matching rows, COUNT() returns 0. COUNT(NULL) returns 0.

(source: https://dev.mysql.com/doc/refman/8.0/en/aggregate-functions.html#function_count)


staabm:
thank you

PR #587, opened 2023-03-23
SQL AST: min/max/avg improvements (#587)Co-authored-by: Markus Staab <maggus.staab@googlemail.com>

### LINKED ISSUE #578: AST: inconsistent behavior with functions and aliases

hemberger:
Fix the avg return type extension using the following rules for its return value:

* `null` if the input is only null.
* `numeric-string` if non-null inputs are only integers.
* `float` if _any_ non-null input is a string.

Put min/max in their own return type extension, since their behavior is sufficiently different from avg. They return the input type, but if any input is a string, then the return value is stringified.

Closes #578.

### DIFF
--- a/src/SqlAst/AvgReturnTypeExtension.php
+++ b/src/SqlAst/AvgReturnTypeExtension.php
@@ -9,6 +9,7 @@ use PHPStan\Type\IntersectionType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\UnionType;
 use SqlFtw\Sql\Expression\BuiltInFunction;
 use SqlFtw\Sql\Expression\FunctionCall;
 
@@ -34,18 +35,32 @@ final class AvgReturnTypeExtension implements QueryFunctionReturnTypeExtension
         }
         $argType = TypeCombinator::removeNull($argType);
 
+        if ($argType instanceof UnionType) {
+            $newTypes = [];
+            foreach ($argType->getTypes() as $type) {
+                $newTypes[] = $this->convertToAvgType($type);
+            }
+            $newType = TypeCombinator::union(...$newTypes);
+        } else {
+            $newType = $this->convertToAvgType($argType);
+        }
+
+        return TypeCombinator::addNull($newType);
+    }
+
+    private function convertToAvgType(Type $argType): Type
+    {
         if ($argType->isInteger()->yes()) {
-            $numberType = new IntersectionType([
+            return new IntersectionType([
                 new StringType(),
                 new AccessoryNumericStringType(),
             ]);
-        } elseif ($argType->isSuperTypeOf(new StringType())->yes()) {
-            $numberType = $argType->toFloat();
-        } else {
-            // We don't know how to handle any other types
-            return null;
         }
 
-        return TypeCombinator::addNull($numberType);
+        if ($argType->isString()->yes()) {
+            return $argType->toFloat();
+        }
+
+        return $argType;
     }
 }
--- a/src/SqlAst/AvgReturnTypeExtension.php
+++ b/src/SqlAst/AvgReturnTypeExtension.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace staabm\PHPStanDba\SqlAst;
 
+use PHPStan\Type\Accessory\AccessoryNumericStringType;
+use PHPStan\Type\IntersectionType;
+use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use SqlFtw\Sql\Expression\BuiltInFunction;
@@ -13,7 +16,7 @@ final class AvgReturnTypeExtension implements QueryFunctionReturnTypeExtension
 {
     public function isFunctionSupported(FunctionCall $expression): bool
     {
-        return \in_array($expression->getFunction()->getName(), [BuiltInFunction::AVG, BuiltInFunction::MIN, BuiltInFunction::MAX], true);
+        return \in_array($expression->getFunction()->getName(), [BuiltInFunction::AVG], true);
     }
 
     public function getReturnType(FunctionCall $expression, QueryScope $scope): ?Type
@@ -25,7 +28,23 @@ final class AvgReturnTypeExtension implements QueryFunctionReturnTypeExtension
         }
 
         $argType = $scope->getType($args[0]);
-        $numberType = $argType->toNumber();
+
+        if ($argType->isNull()->yes()) {
+            return $argType;
+        }
+        $argType = TypeCombinator::removeNull($argType);
+
+        if ($argType->isInteger()->yes()) {
+            $numberType = new IntersectionType([
+                new StringType(),
+                new AccessoryNumericStringType(),
+            ]);
+        } elseif ($argType->isSuperTypeOf(new StringType())->yes()) {
+            $numberType = $argType->toFloat();
+        } else {
+            // We don't know how to handle any other types
+            return null;
+        }
 
         return TypeCombinator::addNull($numberType);
     }
--- a/src/SqlAst/MinMaxReturnTypeExtension.php
+++ b/src/SqlAst/MinMaxReturnTypeExtension.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace staabm\PHPStanDba\SqlAst;
+
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use SqlFtw\Sql\Expression\BuiltInFunction;
+use SqlFtw\Sql\Expression\FunctionCall;
+
+final class MinMaxReturnTypeExtension implements QueryFunctionReturnTypeExtension
+{
+    public function isFunctionSupported(FunctionCall $expression): bool
+    {
+        return \in_array($expression->getFunction()->getName(), [BuiltInFunction::MIN, BuiltInFunction::MAX], true);
+    }
+
+    public function getReturnType(FunctionCall $expression, QueryScope $scope): ?Type
+    {
+        $args = $expression->getArguments();
+
+        if (1 !== \count($args)) {
+            return null;
+        }
+
+        $argType = $scope->getType($args[0]);
+
+        if ($argType->isSuperTypeOf(new StringType())->yes()) {
+            $argType = $argType->toString();
+        }
+
+        return TypeCombinator::addNull($argType);
+    }
+}
--- a/src/SqlAst/MinMaxReturnTypeExtension.php
+++ b/src/SqlAst/MinMaxReturnTypeExtension.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace staabm\PHPStanDba\SqlAst;
 
-use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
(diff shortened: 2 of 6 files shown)


src/SqlAst/MinMaxReturnTypeExtension.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $args = $expression->getArguments();
+
+        if (1 !== \count($args)) {
+            return null;
+        }
+
+        $argType = $scope->getType($args[0]);
+
+        if ($argType->isSuperTypeOf(new StringType())->yes()) {
```
staabm, line 30: isString()->yes()

src/SqlAst/AvgReturnTypeExtension.php
```diff
@@ -25,7 +28,23 @@ public function getReturnType(FunctionCall $expression, QueryScope $scope): ?Typ
         }
 
         $argType = $scope->getType($args[0]);
-        $numberType = $argType->toNumber();
+
+        if ($argType->isNull()->yes()) {
+            return $argType;
+        }
+        $argType = TypeCombinator::removeNull($argType);
```
staabm, line 35: Seems unnecessary?

src/SqlAst/AvgReturnTypeExtension.php
```diff
@@ -25,7 +28,23 @@ public function getReturnType(FunctionCall $expression, QueryScope $scope): ?Typ
...
+
+        if ($argType->isInteger()->yes()) {
+            $numberType = new IntersectionType([
+                new StringType(),
+                new AccessoryNumericStringType(),
+            ]);
+        } elseif ($argType->isSuperTypeOf(new StringType())->yes()) {
```
staabm, line 42: isString()->yes()

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -214,28 +214,53 @@ public function strcase(PDO $pdo): void
         assertType('PDOStatement<array{col: non-empty-string&numeric-string, 0: non-empty-string&numeric-string}>', $stmt);
     }
 
-    public function avgMinMax(PDO $pdo): void
+    public function avg(PDO $pdo): void
     {
         $stmt = $pdo->query('SELECT avg(freigabe1u1) as avg from ada');
-        assertType('PDOStatement<array{avg: int<-32768, 32767>|null, 0: int<-32768, 32767>|null}>', $stmt);
+        assertType('PDOStatement<array{avg: numeric-string|null, 0: numeric-string|null}>', $stmt);
```
staabm, line 220: Is a int really turned into a numeric-string?

Or should it be `int<...>|float` instead?

hemberger: For whatever reason, it really is turned into a `numeric-string`.

```php
$result = $pdo->query('SELECT avg(3)')->fetch();
var_dump($result);
$result = $pdo->query('SELECT avg(account_id) from account')->fetch();
var_dump($result);
```
```
array (size=2)
  'avg(3)' => string '3.0000' (length=6)
  0 => string '3.0000' (length=6)
array (size=2)
  'avg(account_id)' => string '23.0256' (length=7)
  0 => string '23.0256' (length=7)
```

src/SqlAst/AvgReturnTypeExtension.php
hemberger: Removing null is necessary for `int|null` fields to properly pass the `isInteger()` check a couple lines below. Otherwise they incorrectly fall through to the `else` branch. Would you like me to add a comment? (Or is there a better way to do this?)

src/SqlAst/MinMaxReturnTypeExtension.php
hemberger: I believe that `isSuperTypeOf` is necessary here. For example, consider `SELECT max(ifnull(email, adaid)) as max from ada`. The argument to `max` is `string|int`, and if there are _any_ string inputs, the output is always converted to a string (even if the element it selects came from `adaid`).

src/SqlAst/AvgReturnTypeExtension.php
hemberger: I believe that `isSuperTypeOf` is necessary here as well. For example, consider `SELECT avg(ifnull(email, adaid)) as avg from ada`. The argument to `avg` is `string|int`, but PDO will always output the result as a float (even if all `email` are null).

hemberger: Actually, I think the issue with my example above is due to the behavior of `ifnull`, and has nothing to do with avg/min/max. Let me fix this PR and I'll submit a separate PR for ifnull.

The underlying issue was:
> The default return type of [IFNULL(expr1,expr2)](https://dev.mysql.com/doc/refman/8.0/en/flow-control-functions.html#function_ifnull) is the more “general” of the two expressions, in the order STRING, REAL, or INTEGER

src/SqlAst/AvgReturnTypeExtension.php
```diff
@@ -25,8 +29,38 @@ public function getReturnType(FunctionCall $expression, QueryScope $scope): ?Typ
...
+        if ($argType->isInteger()->yes()) {
+            return new IntersectionType([
+                new StringType(),
+                new AccessoryNumericStringType(),
+            ]);
+        }
+
+        if ($argType->isString()->yes()) {
+            return $argType->toFloat();
+        }
```
staabm, line 62: I think this impl deserves test-cases for a "non-empty-string", "non-falsey-string" and "numeric-string"

hemberger: Hmm, I don't know how to coerce a column/field type to be any of those types explicitly, so I'm not sure how I would get those types of inputs to `avg` in this style of test. I know I can get a constant string that is non-empty or numeric, such as `SELECT avg('foo')` or `SELECT avg('123.45')`. Is that what you're looking for? Otherwise, if you have an example, that would be really helpful. Thanks!

src/SqlAst/AvgReturnTypeExtension.php
staabm: I think using CONCAT it should be possible to construct these strings, see https://github.com/staabm/phpstan-dba/blob/6b6e1ca1213fc9bcb4440a946da6b1302297edc1/src/SqlAst/ConcatReturnTypeExtension.php#L65-L76

src/SqlAst/AvgReturnTypeExtension.php
staabm: using CONCAT it should be possible to construct these strings, see https://github.com/staabm/phpstan-dba/blob/6b6e1ca1213fc9bcb4440a946da6b1302297edc1/src/SqlAst/ConcatReturnTypeExtension.php#L65-L76

hemberger: Thanks for the suggestion, this worked great. And just to confirm the expected output types:
```php
$result = $pdo->query('SELECT avg(concat("0"))');
var_dump($result->fetch());
$result = $pdo->query('SELECT avg(concat("foo"))');
var_dump($result->fetch());
```

```
array (size=2)
  'avg(concat("0"))' => float 0
  0 => float 0
array (size=2)
  'avg(concat("foo"))' => float 0
  0 => float 0
```

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -214,28 +214,61 @@ public function strcase(PDO $pdo): void
...
+
+        $stmt = $pdo->query('SELECT avg(coalesce(eladaid, 9999999999999999)) as avg from ak');
+        assertType('PDOStatement<array{avg: numeric-string|null, 0: numeric-string|null}>', $stmt);
+
+        $stmt = $pdo->query('SELECT avg(email) as avg from ada');
+        assertType('PDOStatement<array{avg: float|null, 0: float|null}>', $stmt);
+
+        // Test non-empty-string input
+        $stmt = $pdo->query('SELECT avg(concat("0")) as avg from ada');
```
staabm, line 232: isn't `concat("0") a `numeric-string`?

hemberger: It's both a `numeric-string` and `non-empty-string`:
```
string(39) "SELECT avg(concat("0")) as avg from ada"
object(PHPStan\Type\IntersectionType)#914 (2) {
  ["sortedTypes":"PHPStan\Type\IntersectionType":private]=>
  bool(false)
  ["types":"PHPStan\Type\IntersectionType":private]=>
  array(3) {
    [0]=>
    object(PHPStan\Type\StringType)#896 (0) {
    }
    [1]=>
    object(PHPStan\Type\Accessory\AccessoryNumericStringType)#897 (0) {
    }
    [2]=>
    object(PHPStan\Type\Accessory\AccessoryNonEmptyStringType)#891 (0) {
    }
  }
}
```

staabm: when you use `$type->describe(VerbosityLevel::precise())` it will only print `numeric-string` as all `numeric-string`s are non empty.

I think its impossible to create a non-empty-string with a constant value. you either get a numeric-string or a non-falsey-string :)

lets just fix the comment

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -214,28 +214,61 @@ public function strcase(PDO $pdo): void
...
+        $stmt = $pdo->query('SELECT avg(eladaid) as avg from ak');
+        assertType('PDOStatement<array{avg: numeric-string|null, 0: numeric-string|null}>', $stmt);
+
+        $stmt = $pdo->query('SELECT avg(coalesce(eladaid, 9999999999999999)) as avg from ak');
+        assertType('PDOStatement<array{avg: numeric-string|null, 0: numeric-string|null}>', $stmt);
+
+        $stmt = $pdo->query('SELECT avg(email) as avg from ada');
+        assertType('PDOStatement<array{avg: float|null, 0: float|null}>', $stmt);
+
+        // Test non-empty-string input
```
staabm, line 231:
```suggestion
        // Test numeric-string input
```

staabm:
thank you. great work

staabm:
Sorry, I managed to scramble the commit message somehow

PR #590, opened 2023-03-25
SQL AST: parse expressions with no FROM clause

hemberger:
Expressions without a `FROM` clause are valid SQL, and can be properly parsed by allowing `QueryScope` to take a nullable `Table` argument.

### DIFF
--- a/src/DbSchema/SchemaHasherMysql.php
+++ b/src/DbSchema/SchemaHasherMysql.php
@@ -76,8 +76,10 @@ final class SchemaHasherMysql
 
             try {
                 $result = $this->connection->query($query);
-                $row = $result->fetch_assoc();
-                $hash = $row['dbsignature'] ?? '';
+                if ($result instanceof \mysqli_result) {
+                    $row = $result->fetch_assoc();
+                    $hash = $row['dbsignature'] ?? '';
+                }
             } finally {
                 $this->connection->rollback();
             }
--- a/src/DbSchema/SchemaHasherMysql.php
+++ b/src/DbSchema/SchemaHasherMysql.php
@@ -76,10 +76,8 @@ final class SchemaHasherMysql
 
             try {
                 $result = $this->connection->query($query);
-                if ($result instanceof \mysqli_result) {
-                    $row = $result->fetch_assoc();
-                    $hash = $row['dbsignature'] ?? '';
-                }
+                $row = $result->fetch_assoc();
+                $hash = $row['dbsignature'] ?? '';
             } finally {
                 $this->connection->rollback();
             }
--- a/tests/sqlAst/data/sql-ast-narrowing.php
+++ b/tests/sqlAst/data/sql-ast-narrowing.php
@@ -7,6 +7,15 @@ use function PHPStan\Testing\assertType;
 
 class Foo
 {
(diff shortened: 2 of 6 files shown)


src/DbSchema/SchemaHasherMysql.php
```diff
@@ -76,10 +76,8 @@ public function hashDb(): string
 
             try {
                 $result = $this->connection->query($query);
-                if ($result instanceof \mysqli_result) {
```
staabm, line 79: I think in case of a invalid query or similar, the result will not be a object and we might run into a fatal error after this change?

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -7,6 +7,15 @@
 
 class Foo
 {
+    public function noFromTable(PDO $pdo): void
```
staabm, line 10: does the change also make type inference work for cases where the FROM part is a subselect or similar? if so, would be great to have a test-case for it

src/DbSchema/SchemaHasherMysql.php
hemberger: Without this change, we get a PHPStan warning:

> Instanceof between mysqli_result<array<string, mixed>> and mysqli_result will always evaluate to true.  

My understanding was that this change allows us to parse the query string on line 42, and detects that we can't get a boolean return value from this query.

If you'd prefer to keep the check, I'm happy to add it to the ignored warnings in phpstan.neon.dist instead.

staabm: I think we should leave the check here. I don't see how the query in line 42 is related to the chang regarding queries without a FROM.

in case it really changes the inference for said query, this would mean we miss a new unit test for this scenario

tests/sqlAst/data/sql-ast-narrowing.php
hemberger: No, it does not handle `TableReferenceSubquery` for the FROM clause yet. I kind of see a way to do it by recursively using `QueryReflection::getResultType`, but the `TableReferenceSubquery` class doesn't retain the subquery string, so I think we'd possibly need some changes to SqlFtw to make it happen.

In the meantime, I'll make the logic in this PR more explicit so that it's clear it still does not yet support subqueries.

src/DbSchema/SchemaHasherMysql.php
hemberger: This is fixed (reverted to the original form with the type check) now that we are explicitly handling the case where there is no FROM clause. It was erroneously handling other unsupported FROM clause expressions as if they were the no FROM clause case. Thanks for the careful review!

hemberger:
Instead of allowing `QueryScope` to take a nullable `Table`, I'm now creating an empty `Table` to signify that there is no FROM clause. We could still go the nullable route if that is preferable, but it gets complicated by a bunch of bookkeeping variables to differentiate between `$fromTable === null` because we have an empty FROM clause (allowed) vs. because we have some other non-empty FROM clause that we don't support (disallowed). Another option is to make `Table` non-final and have a few subclasses (which is potentially one way to handle the abstraction for FROM subquery in the future).

staabm:
thank you

PR #592, opened 2023-04-03
SQL AST: ifnull/nullif improvements

hemberger:
Previously the `IfNullReturnTypeExtension` was used for both IFNULL and NULLIF. However, these SQL functions should have different behavior.

`NULLIF(expr1,expr2)`
> Returns NULL if expr1 = expr2 is true, otherwise returns expr1.

`IFNULL(expr1,expr2)`
> If expr1 is not NULL, IFNULL() returns expr1; otherwise it returns expr2.

Note that `IFNULL` has additional type-mutating logic because it has to compare the two values in a temporary table:

> The default return type of IFNULL(expr1,expr2) is the more "general"
> of the two expressions, in the order STRING, REAL, or INTEGER.

We now also require that the functions take exactly two arguments.

The `typemix` table is used for these tests since its column names clearly identify the column type (e.g. `c_tinyint` vs. `gesperrt`).


tests/schema.sql
```diff
@@ -50,6 +50,7 @@ CREATE TABLE `typemix` (
    `c_bit` bit(7) DEFAULT NULL,
    `c_int` int NOT NULL,
    `c_tinyint` tinyint NOT NULL,
+   `c_nullable_tinyint` tinyint,
```
staabm, line 53: please also update the pgsql schema

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -46,44 +46,77 @@ public function coalesce(PDO $pdo): void
...
+        assertType("PDOStatement<array{col: numeric-string, 0: numeric-string}>", $stmt);
+
+        $stmt = $pdo->query('SELECT ifnull(c_nullable_tinyint, "default") as col from typemix');
+        assertType("PDOStatement<array{col: 'default'|numeric-string, 0: 'default'|numeric-string}>", $stmt);
+
+        $stmt = $pdo->query('SELECT ifnull(c_nullable_tinyint, 5000) as col from typemix');
+        assertType('PDOStatement<array{col: 5000|int<-128, 127>, 0: 5000|int<-128, 127>}>', $stmt);
+
+        $stmt = $pdo->query('SELECT ifnull(c_int, c_float) as col from typemix');
+        assertType('PDOStatement<array{col: float, 0: float}>', $stmt);
```
staabm, line 62: wondering why this can't get `int`. `c_int` is not nullable, shouldn't it be `PDOStatement<array{col: int, 0: int}>`?

tests/schema.sql
hemberger: PostgreSQL doesn't have an equivalent of `tinyint`, so I'm not sure what to do here.

tests/sqlAst/data/sql-ast-narrowing.php
hemberger: This is a consequence of the following statement from the documentation:

> The default return type of [IFNULL(expr1,expr2)](https://dev.mysql.com/doc/refman/5.7/en/flow-control-functions.html#function_ifnull) is the more “general” of the two expressions, in the order STRING, REAL, or INTEGER.

See https://dev.mysql.com/doc/refman/5.7/en/flow-control-functions.html#function_ifnull.

Even if `expr1` is not nullable, the result type will still be the type of `expr2` if its type is "more general" than `expr1`. In this test case, the result really will be integer-valued floats:

```
CREATE TEMPORARY TABLE `temp` SELECT IFNULL(c_int, c_float) FROM `typemix`;
DESCRIBE `temp`;
 
Field                     Type
IFNULL(c_int, c_float)	  double
```
And we can confirm this with PDO as well:
```php
$result = $pdo->query('SELECT ifnull(c_int, c_float) as col from typemix');
var_dump($result->fetch(PDO::FETCH_ASSOC));
```
```
array (size=1)
  'col' => float 1
```
It's surprising behavior, to be sure, but it is accurate to the best of my knowledge.

This may be a crazy (and/or bad) idea, but... is it possible to have an `IntersectionType` between `IntegerRangeType` and `FloatType`?

tests/schema.sql
staabm: for pgsql compat it would be better to introduce a different nullable `int` type as `tinyint` then.
at best one which is compatible between mysql and pgsql.

I think we already do some similar things in other tests, to keep them compatible between mysql and pgsql.
just have a look at the [MysqlIntegerRanges ](https://github.com/staabm/phpstan-dba/blob/5d759ac7c57fba21d31284937709ba923f4f8b51/src/Types/MysqlIntegerRanges.php)and [PgsqlIntegerRanges ](https://github.com/staabm/phpstan-dba/blob/5d759ac7c57fba21d31284937709ba923f4f8b51/src/Types/PgsqlIntegerRanges.php)class and find the type which is backend by the same integer range.

that way we don't need to rewrite all tests which use tinyint when pgsql is supported for the SQL AST feature.

tests/sqlAst/data/sql-ast-narrowing.php
staabm: > This may be a crazy (and/or bad) idea, but... is it possible to have an IntersectionType between IntegerRangeType and FloatType?

its not crazy at all. I think this should be the default return type for the IFNULL extension.
we should narrow the type only if we are sure we know better.

src/SqlAst/IfNullReturnTypeExtension.php
```diff
@@ -13,20 +13,78 @@ final class IfNullReturnTypeExtension implements QueryFunctionReturnTypeExtensio
...
+        // If arg1 can be null, the result can be arg1 or arg2;
+        // otherwise, the result can only be arg1.
+        if ($arg1ContainsNull) {
+            $resultType = TypeCombinator::union($argType1NoNull, $argType2NoNull);
+        } else {
+            $resultType = $argType1;
+        }
+
+        // The result type is always the "more general" of the two args
+        // in the order: string, float, integer.
```
staabm, line 49:
```suggestion
        // The result type is always the "more general" of the two args
        // in the order: string, float, integer.
        // see https://dev.mysql.com/doc/refman/5.7/en/flow-control-functions.html#function_ifnull
```

tests/sqlAst/data/sql-ast-narrowing.php
staabm: random idea: maybe we can add some magic to our test-suite (separate PR) which executes the queries against the real database and verifies our asserted-types are really matching those happen at runtime?

hemberger: > its not crazy at all. I think this should be the default return type for the IFNULL extension.

Are you sure this works? When I try to intersect an `IntegerType` with `FloatType` (or numeric-string), I get a `NeverType`. I'm trying:

```php
// Narrow the result type if it can only contain values from arg1
if (! $arg1ContainsNull) {
    $resultType = TypeCombinator::intersect($argType1, $resultType);
}
```

----------------

> random idea: maybe we can add some magic to our test-suite (separate PR) which executes the queries against the real database and verifies our asserted-types are really matching those happen at runtime?

It seems like we could already do this at runtime. We already perform the actual query in `QueryReflection::getResultType`:
```php
$resultType = $reflector->getResultType($queryString, $fetchType);
```
Then after we narrow it with SQL AST, we could make sure the narrowed type is a sub-type of the actual query result? Something like:
```diff
+$originalResultType = $resultType;
 $resultType = $parserInference->narrowResultType($queryString, $resultType);
+if (! $originalResultType->isSuperTypeOf($resultType)->yes()); {
+    throw new ShouldNotHappenException($resultType->describe(VerbosityLevel::precise()) . ' is not a subtype of ' . $originalResultType->describe(VerbosityLevel::precise()));
+}
```
I gave this a quick try, and it seems to properly find cases where SQL AST is returning the improper types. For example, with this SQL AST test of `IF`:
```
$stmt = $pdo->query('SELECT if(freigabe1u1 > 100, "a", 1) as col from ada');
assertType("PDOStatement<array{col: 1|'a', 0: 1|'a'}>", $stmt);
```
we would get `PHPStan\ShouldNotHappenException: array{col: 1|'a', 0: 1|'a'} is not a subtype of array{col: string, 0: string}`, because the actual return type should be `array{col: '1'|'a', 0: '1':'a'}`, because `IF` does the same thing as `IFNULL` where it widens the type of the result set to the more general of the two types, regardless of which arg the result value is taken from.

tests/schema.sql
hemberger: Do we really want to try to shoehorn MySQL types into PostgreSQL tables? As I mentioned in https://github.com/staabm/phpstan-dba/pull/573#issuecomment-1471839303, there are some fundamental differences between the two that cannot be reconciled.

If you really want, I'll add a nullable smallint column and call it `c_nullable_tinyint`, but it's not going to help make the tests reusable since it will be a 2-byte range instead of a 1-byte range.

staabm: ok. lets leave it for now. I created a separate issue with the thing I have in mind

tests/sqlAst/data/sql-ast-narrowing.php
staabm: > Are you sure this works? When I try to intersect an IntegerType with FloatType (or numeric-string), I get a NeverType

sorry, I was talking about a union, not a intersection.

----

> It seems like we could already do this at runtime.

the initial goal of my idea was to verify whether the asserts within the test-suite actually reflect reality.
I don't want to do this checks at runtime of phpstan-dba in the end-user project.
just a mean to harden CI and improve our test-suite.

hemberger: Thanks for all the feedback, it's very much appreciated!

> sorry, I was talking about a union, not a intersection.

It seems to me that a union is not appropriate here since, even though the floats may be integer valued, they will never be integer type. For example, if the user calls `is_int` on the result, it will return false at runtime, but PHPStan would think this is maybe legal. Am I thinking about this the wrong way?

-------------------

> nevertheless: I like your idea of verifying compatibility of the SQL Ast narrowed type to the native type. thats another mean we could use in the testing-job (not at scan time of the end-user project)

I'll think about how we might be able to achieve this in the test suite. It seems like it might be tricky in the `gatherAssertTypes` framework, but maybe we can come up with something workable. Another alternative might be to enable it only in debug mode.

staabm: > It seems to me that a union is not appropriate here since, even though the floats may be integer valued, they will never be integer type. For example, if the user calls `is_int` on the result, it will return false at runtime, but PHPStan would think this is maybe legal. Am I thinking about this the wrong way?

I get it now. its a float then.

------



> I'll think about how we might be able to achieve this in the test suite. It seems like it might be tricky in the `gatherAssertTypes` framework, but maybe we can come up with something workable. Another alternative might be to enable it only in debug mode.

yeah I think we need to play a bit. maybe it can be built using a rule which analyzes the tests... I am not sure yet

staabm:
thanks for the in detail discussion and your analysis. I really appreciate it.

Issue #466, opened 2022-11-16
duplicate stubs error


staabm:
need to figure out a way how we can handle these stub errors and test whether we still need these stub files after all

https://github.com/phpstan/phpstan/issues/8378

I cannot reproduce it yet or know the exact circumstances, but I sometimes get errors like

```
 ------ -----------------------------------------------------------------
  Line   vendor-bin\phpstan\vendor\staabm\phpstan-dba\config\Mysqli.stub
 ------ -----------------------------------------------------------------
  3      Class mysqli_stmt declared multiple times:

 ------ -----------------------------------------------------------------

 ------ ---------------------------------------------------------------------------------------------------------------
  Line   vendor-bin\phpstan\vendor\staabm\phpstan-dba\config\PdoStatement.stub
 ------ ---------------------------------------------------------------------------------------------------------------
  11     Class PDOStatement declared multiple times:
         -
         phar://C:/dvl/Workspace/daiber/vendor-bin/phpstan/vendor/phpstan/phpstan/phpstan.phar/stubs/PDOStatement.stub
         :8
 ------ ---------------------------------------------------------------------------------------------------------------

 ------ ----------------------------------------------------------------------------------------------------
  Line   vendor-bin/phpstan/vendor/phpstan/phpstan/phpstan.phar/stubs/PDOStatement.stub
 ------ ----------------------------------------------------------------------------------------------------
  8      Class PDOStatement declared multiple times:
         - C:/dvl/Workspace/daiber/vendor-bin/phpstan/vendor/staabm/phpstan-dba/config/PdoStatement.stub:11
 ------ ----------------------------------------------------------------------------------------------------

 ------ --------------------------------------------------------------------------
  Line   vendor-bin/phpstan/vendor/phpstan/phpstan/phpstan.phar/stubs/mysqli.stub
 ------ --------------------------------------------------------------------------
  3      Class mysqli_stmt declared multiple times:

 ------ --------------------------------------------------------------------------

```

staabm:
see https://github.com/phpstan/phpstan-src/pull/2096

pscheit:
@staabm since the PdoStatement.stub is now in phpstan-src merged + released, it needs to be removed here, too, right?

staabm:
It doesn't work yet with stubs from phpstan-src only. 


I don't know yet how we can make this error go away

pscheit:
since we have loaded bleedingEdge.neon I (for now) disabled the feature that detects doubles:

```neon
[...]

PR #594, opened 2023-04-03
remove PdoStatement stub

staabm:
refs https://github.com/staabm/phpstan-dba/issues/466


staabm:
it seems one solution could be to use `iterable<array{email: string, id: int}>&PDOStatement` instead of `PDOStatement<array{email: string, id: int}>`

PR #596, opened 2023-04-09
SQL AST: narrow result types with WHERE condition

hemberger:
The WHERE condition of a SQL statement may narrow the type returned from a table. We can account for this when getting the type of each selected column.

Currently, the only implemented conditions are `IS NOT NULL` and `IS NULL`. This can be expanded in the future to support other binary operators and comparison operators.

Since the WHERE condition can narrow selections that are implicitly specified with an asterisk, we now expand the asterisk to an explicit list of columns so that we can perform the AST-based type resolution on them.

### DIFF
--- a/tests/sqlAst/data/sql-ast-narrowing.php
+++ b/tests/sqlAst/data/sql-ast-narrowing.php
@@ -515,6 +515,15 @@ class Foo
         assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, eadavk: numeric-string|null, 1: numeric-string|null}>', $stmt);
     }
 
+    public function joinWhereCondition(PDO $pdo): void
+    {
+        $stmt = $pdo->query('SELECT c_json FROM typemix LEFT JOIN ada ON (c_int = adaid) WHERE c_json IS NOT NULL');
+        assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);
+
+        $stmt = $pdo->query('SELECT c_json FROM typemix RIGHT JOIN ada ON (c_int = adaid) WHERE c_json IS NOT NULL');
+        assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);
+    }
+
     public function multipleJoins(PDO $pdo): void
     {
         $stmt = $pdo->query('SELECT adaid, eladaid, c_int, c_char5 from ada inner join ak on adaid = eladaid inner join typemix on adaid = c_int');
--- a/tests/sqlAst/data/sql-ast-narrowing.php
+++ b/tests/sqlAst/data/sql-ast-narrowing.php
@@ -522,6 +522,12 @@ class Foo
 
         $stmt = $pdo->query('SELECT c_json FROM typemix RIGHT JOIN ada ON (c_int = adaid) WHERE c_json IS NOT NULL');
         assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);
+
+        $stmt = $pdo->query('SELECT c_json FROM ada LEFT JOIN typemix ON (c_json = c_json) WHERE c_json IS NOT NULL');
+        assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);
+
+        $stmt = $pdo->query('SELECT c_json FROM ada RIGHT JOIN typemix ON (c_json = c_json) WHERE c_json IS NOT NULL');
+        assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);
     }
 
     public function multipleJoins(PDO $pdo): void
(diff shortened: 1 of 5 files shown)


tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -7,6 +7,61 @@
 
 class Foo
 {
+    public function whereIsNotNull(PDO $pdo): void
```
staabm, line 10: please add tests which show that a WHERE from a subselect does not affect the outer selects select-columns

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -7,6 +7,61 @@
...
+        // selected with an alias
+        $stmt = $pdo->query('SELECT c_json as col FROM typemix WHERE c_json IS NOT NULL');
+        assertType('PDOStatement<array{col: string, 0: string}>', $stmt);
+
+        // affects input to a function
+        $stmt = $pdo->query('SELECT ifnull("1", c_json) as col FROM typemix WHERE c_json IS NOT NULL');
+        assertType('PDOStatement<array{col: string, 0: string}>', $stmt);
+
+        // selected with an asterisk
+        $stmt = $pdo->query('SELECT * FROM typemix WHERE c_json IS NOT NULL');
```
staabm, line 27: another case which can be imlpemented in a separate PR:
`'SELECT nullable-col FROM typemix WHERE non-nullable-col = nullable-col'`
assert select of nullable-col cannot contain null

staabm (commented):
looks great

staabm (approved):
great job. this can land after rebase.

thank you

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -7,6 +7,77 @@
 
 class Foo
 {
+    public function whereIsNotNull(PDO $pdo): void
```
staabm, line 10: another thing just came to mind: I think we need another test which asserts the type of a column from a joined table narrowed via WHERE

hemberger: Good suggestion, I've rebased and added tests for this case. Thanks!

tests/sqlAst/data/sql-ast-narrowing.php
```diff
@@ -440,6 +515,15 @@ public function joinSelectOutsideCondition(PDO $pdo): void
         assertType('PDOStatement<array{adaid: int<-32768, 32767>, 0: int<-32768, 32767>, eadavk: numeric-string|null, 1: numeric-string|null}>', $stmt);
     }
 
+    public function joinWhereCondition(PDO $pdo): void
+    {
+        $stmt = $pdo->query('SELECT c_json FROM typemix LEFT JOIN ada ON (c_int = adaid) WHERE c_json IS NOT NULL');
+        assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);
+
+        $stmt = $pdo->query('SELECT c_json FROM typemix RIGHT JOIN ada ON (c_int = adaid) WHERE c_json IS NOT NULL');
```
staabm, line 523: Does this PR also support the case when the selected column which is narrowed via WHERE is from a joined table?
(Another test)

hemberger: Yes, it does. However, there's currently a bug in `QueryScope::narrowJoinedColumnType` where if the selected column is from the joined table but not in the join condition, then the AST type resolution returns `MixedType`, and so we revert to the query simulation type resolution (i.e. the WHERE condition is ignored, since query simulation doesn't narrow types like this).

Therefore, the following tests that I was planning to add would fail, because we get `string|null` (the query simulation type) instead of `string` (the expected AST type).
```php
$stmt = $pdo->query('SELECT c_json FROM ada LEFT JOIN typemix ON (c_int = adaid) WHERE c_json IS NOT NULL');
assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);

$stmt = $pdo->query('SELECT c_json FROM ada RIGHT JOIN typemix ON (c_int = adaid) WHERE c_json IS NOT NULL');
assertType('PDOStatement<array{c_json: string, 0: string}>', $stmt);
```
As a workaround, I've added `c_json` to the join condition in an unrealistic way (`c_json = c_json`) since no other table has a nullable string type, and I didn't want to make more schema changes (comparing against a non-nullable string column would prevent us from properly checking that the WHERE condition was used). This allows the AST type resolution to succeed, and we get the column from the joined table narrowed by the WHERE condition.

I'm working on a PR for the table join bug, but that won't be done until later this week I think.

staabm: Cool thanks. Lets look at more cases in a separate PR. 

The PR as is - is already very valuable

PR #608, opened 2023-07-06
Make query in SchemaHashMysql independent from internal sorting

staabm:


src/DbSchema/SchemaHasherMysql.php
```diff
@@ -43,19 +43,24 @@ public function hashDb(): string
...
+                        COALESCE(COLUMN_NAME, ""),
+                        COALESCE(EXTRA, ""),
+                        COLUMN_TYPE,
+                        IS_NULLABLE
+                    ) as columns
+                FROM
+                    information_schema.columns
+                WHERE
+                    table_schema = DATABASE()
+                ORDER BY columns
```
staabm, line 62: table, columns?

PR #613, opened 2023-07-29
Add a real matrix to Static analysis workflow

szepeviktor:
- Added a real 3×2×1×1 matrix + 3 "includes"
- Move database options to `db-options`
- PHP 8.0 + MariaDB was excluded - Should I exclude it now?

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -22,10 +22,10 @@ jobs:
         db-image:
           - "mysql:8.0"
           - "mariadb:latest"
-        mode:
-          - "recording"
         reflector:
           - "mysqli"
+        mode:
+          - "recording"
 
         include:
           # PHPStan development version
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -19,7 +19,6 @@ jobs:
           - "8.0"
           - "8.1"
           - "8.2"
-          - "8.3"
         db-image:
           - "mysql:8.0"
           - "mariadb:latest"
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -19,6 +19,7 @@ jobs:
           - "8.0"
           - "8.1"
           - "8.2"
+          - "8.3"
         db-image:
           - "mysql:8.0"
           - "mariadb:latest"
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -28,9 +28,16 @@ jobs:
           - "recording"
 
         include:
+          - db-image: "mysql:8.0"
+            db-options: '--health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3'
+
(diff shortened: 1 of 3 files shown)


szepeviktor:
@staabm You may copy it to the unit tests workflow.

staabm:
> PHP 8.0 + MariaDB was excluded - Should I exclude it now?

If green - we can keep them

---


Could you have a look at the failling jobs?

.github/workflows/phpstan.yml
```diff
@@ -60,13 +53,13 @@ jobs:
...
     services:
-      mysql:
+      database:
         image: ${{ matrix.db-image }}
         env:
           MYSQL_ROOT_PASSWORD: root
         ports:
           - 3306:3306
-        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+        options: ${{ startsWith(matrix.db-image, 'mariadb') && '--health-cmd="healthcheck.sh --connect --innodb_initialized" --health-interval=1s --health-timeout=10s --health-retries=60' || '--health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3' }}
```
staabm, line 62: can we move the command into the matrix, where we define the db image?

szepeviktor:
> Could you have a look at the failling jobs?

Those job are in the unit test workflow. I leave them 101% up to you.

.github/workflows/phpstan.yml
szepeviktor: Yes ❗

szepeviktor: Done 🍏

image

.github/workflows/phpstan.yml
```diff
@@ -19,6 +19,7 @@ jobs:
           - "8.0"
           - "8.1"
           - "8.2"
+          - "8.3"
```
szepeviktor, line 22: This is too daring 👴🏻 for viktor.

staabm: I agree. I just want to see how the matrix explodes with this additional version, which we need in the near future. will revert afterwards

szepeviktor:
```
tomasvotruba/unused-public 0.2.0.72 requires php ^7.2 || 8.0.*
 -> your php version (8.3.0-dev) does not satisfy that requirement.
```

staabm: I guess this will fix it: https://github.com/staabm/phpstan-dba/pull/619

szepeviktor:
@TomasVotruba uses this mysterious `.72` tail on his version.

```
tomasvotruba/unused-public 0.2.0.72 requires php ^7.2 || 8.0.*
 -> your php version (8.3.0-dev) does not satisfy that requirement.
```

How to get the `.72`-less version?

staabm:
I think the problem is that there currently is no nette/utils release which supports PHP8.3

https://github.com/nette/utils/pull/297

staabm:
thank you

PR #618, opened 2023-07-30
Fix MariaDB health check

staabm:
inspired by https://github.com/staabm/phpstan-dba/pull/613

extracted the parts which actually fix the current failling builds.
refactoring the matrix, like done in #613 should be a separate PR.

Co-authored-by: @szepeviktor

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -18,39 +18,47 @@ jobs:
         include:
           - php-version: "8.0"
             db-image: 'mysql:8.0'
+            db-options: '--health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3'
             reflector: "pdo-mysql"
             mode: "recording"
             phpstan-version: "phpstan@dev"
 
           - php-version: "8.0"
             db-image: 'mysql:8.0'
+            db-options: '--health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3'
             reflector: "pdo-mysql"
             mode: "recording"
           - php-version: "8.0"
             db-image: 'mysql:8.0'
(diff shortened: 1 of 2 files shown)


.github/workflows/phpstan.yml
```diff
@@ -60,13 +60,13 @@ jobs:
...
     services:
-      mysql:
+      database:
         image: ${{ matrix.db-image }}
         env:
           MYSQL_ROOT_PASSWORD: root
         ports:
           - 3306:3306
-        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+        options: ${{ startsWith(matrix.db-image, 'mariadb') && '--health-cmd="healthcheck.sh --connect --innodb_initialized" --health-interval=1s --health-timeout=10s --health-retries=60' || '--health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3' }}
```
staabm, line 69: reverted to the variant I did not like at first, because its the easiest way for now to get the builds green - which is most important.

staabm:
thanks @szepeviktor

.github/workflows/phpstan.yml
szepeviktor: May I disagree?
The most important is long-term sustainability.
Why fix an error today if you have to refix it 12 more times in the coming months??

szepeviktor:
Glad to contribute.

.github/workflows/phpstan.yml
staabm: I had a red CI pipeline for mariadb for a few months and did not know why. thanks to your PR I have an idea how to fix, so I want at first fix the problem to get the pipeline green again, which is super important.

I am still open to the matrix refactoring PR you submitted

PR #621, opened 2023-07-30
Fix Composer usage in Unit Tests MySQL workflow

szepeviktor:
Just using Composer the right way.


szepeviktor:
> [!NOTE]  
> I let you build that matrix. It is fun 🎈

szepeviktor:
BTW prefer-dist and prefer-lowest are totally different animals.
Here it is: https://gist.github.com/szepeviktor/200b45f9805dc8bf3f185f0ade216987

staabm:
thanks

szepeviktor:
Glad to contribute.

PR #630, opened 2023-09-29
Fix false-positive parameter validation error when query string is not resolvable

staabm:


src/QueryReflection/PlaceholderValidation.php
```diff
@@ -36,8 +40,8 @@ public function checkQuery(Expr $queryExpr, Scope $scope, array $parameters): it
 
         $minPlaceholderCount = PHP_INT_MAX;
         $maxPlaceholderCount = 0;
-        foreach ($queryStrings as $queryString) {
-            $placeholderCount = $queryReflection->countPlaceholders($queryString);
+        foreach ($queryStrings as $unnamedQueryString) {
```
staabm, line 43: variable rename required because of https://github.com/phpstan/phpstan/issues/9940

PR #638, opened 2024-01-21
FetchAll actually returns list of results, not an array per se.

patrickkusebauch:


src/Rules/SyntaxErrorInDibiPreparedStatementMethodRule.php
```diff
@@ -112,13 +112,7 @@ private function checkErrors(CallLike $callLike, Scope $scope, MethodReflection
             $parameterType = $scope->getType($parameterExpr);
 
             if ($parameterType instanceof StringType) {
-                $resolvedString = $queryReflection->resolveQueryString($parameterExpr, $scope);
```
staabm, line 115: looks like a unrelated change?

patrickkusebauch: Well, the original method `resolveQueryString` is deprecated, the suggested replacement `resolveQueryStrings` works, but the pass-through of the deprecated method is to a different method that is not doing what is expected and therefore is broken in this use case.

Basically, fixed something that was broken and would not allow any testing to pass in the first place, but was too lazy to do it in a separate stacked PR:

staabm: I will move it into a separate PR, thank you

staabm:
fyi, this will only work if the list-type is enabled on your projects, which means e.g. bleeding-edge

patrickkusebauch:
Yeah, that's fair.

Issue #642, opened 2024-01-27
dibi support DATE_FORMAT


patrickkusebauch:
Currently the `DibiReflection` bails out when it encounters the `%m` modifier anywhere in the query. However, this modifier is a valid modifier for `DATE_FORMAT` SQL function and in such case, it is not replaced by Dibi query builder and it should be just kept and the query reflection should not bail out.

PR #644, opened 2024-02-01
Update DibiReflection.php to replace date_format

patrickkusebauch:
Closes: https://github.com/staabm/phpstan-dba/issues/642

It is a naive solution, and certainly does not cover all the possibilities, but it is a step in the right direction, IMHO.


staabm:
thanks for the PR. its not acceptable without a test, which fails without the fix

PR #651, opened 2024-05-16
Support nested quotes

b-viguier:
Hi 👋 🙂 

I encountered an edge case with the Regex currently used to parse placeholders.

```
payload ->> '$."value-1"' = :value1
OR
payload ->> '$."value-2"' = :value2
```

ℹ️  `"` are needed here to escape the `-` special character in the field's name

The problem with the current `(["\'])([^"\']*\1)` regex is that it tries to exclude `'` **AND** `"` in the middle of a _string_. If this is a simply quoted string we should exclude only simple quotes. 

I found a way to negate a backreference in [stackoverflow](https://stackoverflow.com/a/8057827).

~Unfortunately, I totally failed to run tests locally, so I didn't add corresponding test... but I would love to if you have some helping instructions~ 
~[Edit] it seems that locally I got the same errors encountered by the CI, so maybe the issue is not related to my local setup but to `master` branch...~
[Edit 2 ] tests added 🎉  

Thanks for you work 🙏

### DIFF
--- a/.phpstan-dba-mysqli.cache
+++ b/.phpstan-dba-mysqli.cache
@@ -1,280 +1,757 @@
 <?php return array (
   'schemaVersion' => 'v11-phpstan1_9_3-update',
-  'schemaHash' => 'f8243f08e312cc01a89e4e0e639e78ef',
+  'schemaHash' => '72813316a311d2e2b9384c458247d011',
   'records' => 
   array (
-    'SELECT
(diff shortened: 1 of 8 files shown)


staabm:
Just fixed the build with https://github.com/staabm/phpstan-dba/pull/652

please try again while making sure you are running the tests with `doctrine/dbal:^3` like I did in above PR

staabm:
Thank you!

Issue #653, opened 2024-05-16
CI: Separate Testing of Doctrine 3.x and 4.x


staabm:
refs https://github.com/staabm/phpstan-dba/pull/652  - PRs welcome

- setup a CI testing matrix for doctrine 3.x/4.x)
- separate tests where 3.x and 4.x are not compatible and skip tests which only work for certain versions

clxmstaab:
implemented in https://github.com/staabm/phpstan-dba/pull/754

PR #654, opened 2024-05-16
Fix composer scripts for doctrine/dbal 3.x

staabm:
followup to https://github.com/staabm/phpstan-dba/pull/652

refs https://github.com/staabm/phpstan-dba/issues/653

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,7 +106,7 @@
 		],
 		"clean": [
 			"composer remove sqlftw/sqlftw --ignore-platform-req=php+",
-			"composer remove doctrine/dbal:^3 --ignore-platform-req=php+"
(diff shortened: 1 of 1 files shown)


staabm:
depends on https://github.com/composer/composer/issues/11986

composer.json
```diff
@@ -102,7 +102,7 @@
 		],
 		"build": [
 			"composer require sqlftw/sqlftw --ignore-platform-req=php+",
-			"composer require doctrine/dbal --ignore-platform-req=php+"
+			"composer require doctrine/dbal:^3 --ignore-platform-req=php+"
```
staabm, line 105: use notation which does not require windows-specific escaping

```suggestion
			"composer require doctrine/dbal:3.* --ignore-platform-req=php+"
```

PR #669, opened 2024-08-02
Fix getting columns when sql mode is using ansi_quotes

### LINKED ISSUE #668: Querying the column information fails when sql mode is set to ANSI_QUOTES
I get an error at `src/QueryReflection/PdoMysqlQueryReflector.php:100`: `Column not found: 1054 Unknown column '' in 'field list'`. I can reproduce the same error when running that query for the information schema separate from phpstan.

I found if I remove the `coalesce` functions from that query it seems to work.

My PDO setup runs `SET sql_mode = ANSI_QUOTES` (amongst others). If I remove that one, the query also works.

This was introduced in [`47e28b8` (#261)](https://github.com/staabm/phpstan-dba/pull/261/commits/47e28b8d0af1eb6bceb3106ff405caf2c577f21d). I'm not running MariaDB, but apparently this query trips on my setup.

lode:
This fixes #668.


staabm:
Thanks for the PR. Would be great to have a unit test for the actual problem

lode:
> Thanks for the PR. Would be great to have a unit test for the actual problem

I agree that would be nice!

I tried but couldn't really get this to work. (Various things, didn't have the php extensions installed, tried putting it in Docker, that worked, but about half the existing tests fail, don't fully understand the structure of the existing tests.)

I did add the structure of what I think would be the test.

Do you have some guidelines of how to proceed?

Otherwise, feel free to take my code and make it work if you do have more concrete knowledge about the test setup.

staabm:
I will have a look into a test when time allows, thanks for the fix.

Issue #667, opened 2024-08-02
Enabling `staabm/phpstan-dba` significantly increases garbage collection activity


glaubinix:
We noticed that since we enabled `staabm/phpstan-dba` in our project, a full PHPStan run without cache takes significantly longer than without it.

Running it with the extension
```
php ./vendor/bin/phpstan --debug  537.54s user 3.92s system 97% cpu 9:13.27 total
```
Running it without the extension
```
php ./bin/phpstan --debug  230.65s user 3.89s system 97% cpu 3:59.47 total
```

I triggered a trace with Tideways to see why this suddenly takes so much longer and it turns out that about 6min of the run is spent on garbage collection. Turning garbage collection off results in a significant performance boost.

Running it with the extension
```
php -d zend.enable_gc=0 ./vendor/bin/phpstan --debug  176.72s user 4.04s system 96% cpu 3:06.44 total
```
Running it without the extension
```
php -d zend.enable_gc=0 ./vendor/bin/phpstan --debug  153.03s user 3.25s system 98% cpu 2:39.39 total
```

Is there any chance you encountered similar effects and already have an idea what could cause this? Otherwise, I'll look into some memory profiling in the next days to see if I can find a cause for this. Happy to share additional info but not really sure what might actually be helpful here.

staabm:
Thanks for the report.

Is the problem reproducible in a minimal repo or similat?

glaubinix:
As of right now, I haven't been able to reproduce this on smaller repositories. For instance, running PHPStan just on certain subfolders where we have our SQL queries, runs through in a few seconds and garbage collection activity is similar with and without the extension enabled.

glaubinix:
The config is more or less the default. Setting debug false/true doesn't make a difference

```
<?php declare(strict_types=1);

use staabm\PHPStanDba\QueryReflection\PdoPgSqlQueryReflector;
use staabm\PHPStanDba\QueryReflection\QueryReflection;
use staabm\PHPStanDba\QueryReflection\RecordingQueryReflector;
use staabm\PHPStanDba\QueryReflection\ReflectionCache;
use staabm\PHPStanDba\QueryReflection\RuntimeConfiguration;

QueryReflection::setupReflector(
    new RecordingQueryReflector(
        ReflectionCache::create(
            __DIR__ . '/../.phpstan-dba.cache',
        ),
        new PdoPgSqlQueryReflector(
            new \PDO('pgsql:host=localhost;port=5432;dbname=database'),
[...]

PR #670, opened 2024-08-03
Reduce use of PreviousConnectingVisitor:: ATTRIBUTE_PARENT to assist garbage collector

staabm:
idea is to have less Ast\Nodes with a pointer to the parent node, so the garbage collector has a easier job identifying unused nodes and clear memory. for our use-case its enough to know parents within functions/method bodies.

refs https://github.com/staabm/phpstan-dba/issues/667

---

in the long run it would be best to get rid of PreviousConnectingVisitor but I am currently not deep enough into the details of this lib todo such refactoring right now

### DIFF
--- a/src/Ast/ExpressionFinder.php
+++ b/src/Ast/ExpressionFinder.php
@@ -151,10 +151,6 @@ final class ExpressionFinder
         }
 
         $parent = $node->getAttribute(PreviousConnectingVisitor::ATTRIBUTE_PARENT);
-        if ($parent instanceof FunctionLike) {
-            return null;
-        }
-
         if ($parent instanceof Node) {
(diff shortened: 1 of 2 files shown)


glaubinix:
Thank you for looking into this already. I run latest stable and this branch with GC enabled and disabled. Looks like there is a slight improvement when gc is disabled. Unfortunately, the regular run got a lot worse.

|      |php ./bin/phpstan|php -d zend.enable_gc=0 ./bin/phpstan|
|----|--------|-----------------|
|0.2.8.1|9:04.60| 3:01.07|
|less-parents |13:31.71| 2:57.27 |

All good if there is no easy solution available at the moment. We can disable gc in our CI build for now. If you ever need to let me run anything then just let me know!

PR #709, opened 2024-11-11
PHPStan 2.0 support

staabm:


staabm:
blocked on https://github.com/phpstan/phpstan/discussions/12046 as we are currently not able to read the reflection-cache back into regular PHPStan Type objects

Issue #737, opened 2025-01-29
unsigned int recognized as type of int<-2147483648, 2147483647>


danrot:
I have the following table structure (output from a `SHOW CREATE TABLE` statement in MySQL and omitted fields that do not matter for the issue):

```sql
Create Table: CREATE TABLE `t_documents` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`),
) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8 ROW_FORMAT=DYNAMIC
```

When I query that table in my code, then phpstan-dba will result in the following type:

```
array{id: int<-2147483648, 2147483647>}
```

Which I think is wrong, since the minimum value should be `0`, because the `id` column is defined as `unsigned`.

I wanted to update the tests in a way that show that issue, but it took me too long... I am still willing to contribute a test and maybe even the fix, but I wanted to make sure I am not making any stupid mistakes 😅

staabm:
thanks for reporting. which database driver are you using? I guess the problem only happens for certain drivers (pdo, mysqli, doctrine/dbal,..?)

danrot:
In this project we are using Doctrine DBAL, cannot tell if it also happens with other drivers as well right now.

uuf6429:
Just thinking out loud, could it make sense to have a `TypemixDoctrineDbal` test? Essentially a [`TypemixMysql`](https://github.com/staabm/phpstan-dba/blob/main/tests/default/data/typemix-mysql.php) with for DoctrineDbal. Or is a case-by-case test preferred?

staabm:
sounds like a good idea

PR #738, opened 2025-01-29
Change schema to enforce error

danrot:


tests/schema.sql
```diff
@@ -30,7 +30,7 @@ ALTER TABLE `ak`
 
 
 CREATE TABLE `typemix` (
-   `pid` int NOT NULL,
+   `pid` int unsigned NOT NULL AUTO_INCREMENT,
```
staabm, line 33: I would prefer adding a new column at the end of this table instead of changing a existing column. 

Additionally we need the same schema change in the pgsql schema file

danrot: I am still trying to reproduce my issue https://github.com/staabm/phpstan-dba/issues/737... At first I thought it was somehow related to the column being the primary key, for which reason I added it here. And even though there seems to be some inconsistencies here since it is already breaking some tests, it's not failing in the way I supposed it to 🙈 I think I have to continue to try reproducing this...

PR #740, opened 2025-01-30
Remove redundant test and constants

### LINKED ISSUE #739: MySQL test inside PostgreSQL test class
I'm just wondering if the following method (with respect to the containing class/file) is maybe a copy & paste mistake?
https://github.com/staabm/phpstan-dba/blob/525a9fae5d03f482ada9dd31d1ce039a11565a2d/tests/default/data/typemix-pgsql.php#L13

At one point I thought maybe it was meant to test `pg_` functions, but that driver doesn't seem to be supported at all.

I can't otherwise see any connection between postgres and mysql. 🤔

uuf6429:
Fixes #739.

### DIFF
--- a/tests/default/data/typemix-mysql.php
+++ b/tests/default/data/typemix-mysql.php
@@ -8,8 +8,6 @@ use function PHPStan\Testing\assertType;
 
 class TypemixMysql
 {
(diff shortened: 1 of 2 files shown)


staabm:
//cc @hemberger could you double check we did not oversee something?

staabm:
just found https://github.com/staabm/phpstan-dba/commit/be9fc1f9a5250e5533458c54e6925dd1388dbf5d in which I have removed the constant private array references to get more detailed results, but forgot to drop the constants declaration

staabm:
thank you

PR #766, opened 2025-08-06
Add support for `StaticCall`

### LINKED ISSUE #765: Support for static method calls for queries
I have a little wrapper class for PDO queries that basically boils down to this:
```php
<?php
class DB
{
	public static function Instance() : \Pdo\Mysql
	{
		return PDO::connect();
	}

	/**
	 * @param literal-string $query
	 *
	 * @see \PDO::prepare()
	 */
	public static function Prepare( string $query ) : \PDOStatement
	{
		return self::Instance()->prepare( $query );
	}

	/**
	 * @param literal-string $query
	 *
	 * @see \PDO::query()
	 */
	public static function Query( string $query ) : \PDOStatement
	{
		return self::Instance()->query( $query );
	}
}
```

I tried the following config:
```neon
services:
	-
		class: staabm\PHPStanDba\Rules\SyntaxErrorInQueryMethodRule
		tags: [phpstan.rules.rule]
		arguments:
			classMethods:
				- 'DB::Query#0'
				- 'DB::Prepare#0'
	-
		class: staabm\PHPStanDba\Rules\QueryPlanAnalyzerRule
		tags: [phpstan.rules.rule]
		arguments:
			classMethods:
				- 'DB::Query#0'
				- 'DB::Prepare#0'
```

But after looking at these rules, they only implement `MethodCall`, while I believe static methods would be `StaticCall`, unfortunately `SyntaxErrorInQueryFunctionRule` also does not work for this.

Is the solution here to add a 3rd rule to handle static calls, and then another if check in QueryPlanAnalyzerRule?

xPaw:
Fixes #765

I am unable to get tests working locally, so I can't really add one.


staabm:
without a test we cannot merge the PR. how do the problems locally manifest?

xPaw:
For some reason running composer test tries to install some packages, including adding sqlftw/sqlftw to composer, and then just quits because it can't find phpstan-dba\tests\default\config/../../..\.env

staabm:
> or at least was documented in a contributing guide or something.

https://github.com/staabm/phpstan-dba/commit/a5cee1809325b85f9e0687031249012d68995ad0

xPaw:
So there's skipped tests that use `getenv()` checks, for some reason the .env loaded variables are only available in `$_ENV`.

tests/rules/data/query-plan-analyzer.php
```diff
@@ -72,4 +73,10 @@ public function bug442(Connection $conn, string $table)
         $query = "SELECT email, adaid FROM `$table` WHERE adaid = ?";
         $conn->fetchAssociative($query, [1]);
     }
+
+    public function staticCallTests(): void
+    {
+        StaticDatabase::query("SELECT * FROM `ada` WHERE email = 'test@example.com';");
+        StaticDatabase::executeQuery("SELECT *,adaid FROM `ada` WHERE email = 'test@example.com';");
```
staabm, line 80: to reduce the changed files, you could do a fully-qualified call here

```suggestion
        staabm\PHPStanDba\Tests\Fixture\StaticDatabase::query("SELECT * FROM `ada` WHERE email = 'test@example.com';");
        staabm\PHPStanDba\Tests\Fixture\StaticDatabase::executeQuery("SELECT *,adaid FROM `ada` WHERE email = 'test@example.com';");
```

and remove the namespace import

tests/rules/data/syntax-error-in-prepared-statement.php
```diff
@@ -382,4 +383,10 @@ protected function conditionalUnnamedPlaceholder(Connection $connection, array $
 
         $connection->preparedQuery($query, $values);
     }
+
+    public function staticCallPreparedStatementTests(): void
+    {
+        StaticDatabase::preparedQuery('SELECT email adaid WHERE gesperrt freigabe1u1 FROM ada', []);
+        StaticDatabase::prepare('SELECT email, adaid, gesperrt, freigabe1u1 FROM ada WHERE asdsa=?', [1]);
```
staabm, line 390: same here

tests/rules/QueryPlanAnalyzerRuleTest.php
```diff
@@ -54,14 +57,6 @@ public static function getAdditionalConfigFiles(): array
 
     public function testNotUsingIndex(): void
     {
-        if ('pdo-pgsql' === getenv('DBA_REFLECTOR')) {
```
xPaw, line 57: didnt mean to commit this

xPaw:
composer.json changes every time I run composer test by the way.

staabm:
> composer.json changes every time I run composer test by the way.

yeah.. thats a sideffect of running the tests. its not expected to be committed

staabm:
thank you. released in 0.4.2

Issue #770, opened 2025-08-17
PDOStatement with bind calls does not resolve types on fetch


xPaw:
Test case:
```php
    public function testWithBind(PDO $pdo)
    {
        $stmt = $pdo->prepare('SELECT email, adaid FROM ada WHERE adaid = :adaid');
        $stmt->bindValue('adaid', 1);
        $stmt->execute();
        $all = $stmt->fetch(PDO::FETCH_ASSOC);
        assertType('array{email: string, adaid: int<-32768, 32767>}|false', $all);
    }
```

```diff
-'array{email: string, adaid: int<-32768, 32767>}|false'
+'mixed'
```

I've roughly boiled it down to `PdoMysqlQueryReflector` not accepting parameters into `simulateQuery` which means the query just throws on unbound parameters. Even though this PDOException gets cached, it never gets reported anywhere, so I didn't even know if this was an issue with the PDO calls not being discovered or not.

`PdoStatementExecuteMethodRule` does try to find and resolve the bind parameters, so I wonder if this should actually attach the type to the statement somehow, so that `fetch` calls didn't need to simulate the query again.

This also means `PdoPrepareDynamicReturnTypeExtension` tries to simulate the query with placeholders and gets PDOException due to unbound parameters.

Mildly related to #275.

PR #772, opened 2026-04-08
Fix mixed row inference for PDOStatement::bindValue() + execute()

vaielab:
## Summary

Fix row inference for PDO statements that use `bindValue()` before `execute()`.

The issue caused iterated rows to be inferred as `mixed` instead of the expected typed array shape.

## Changes

- add regression coverage for the `bindValue()` path
- update `PdoStatementExecuteTypeSpecifyingExtension` to preserve row typing after `execute()`

## Result

`foreach ($stmt as $row)` now keeps the expected inferred array type for supported `bindValue()` usages.

### DIFF
--- a/src/Extensions/PdoStatementExecuteTypeSpecifyingExtension.php
+++ b/src/Extensions/PdoStatementExecuteTypeSpecifyingExtension.php
@@ -12,6 +12,9 @@ use PHPStan\Analyser\TypeSpecifier;
 use PHPStan\Analyser\TypeSpecifierAwareExtension;
 use PHPStan\Analyser\TypeSpecifierContext;
 use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantIntegerType;
+use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\MethodTypeSpecifyingExtension;
 use PHPStan\Type\Type;
 use staabm\PHPStanDba\PdoReflection\PdoStatementReflection;
@@ -62,9 +65,7 @@ final class PdoStatementExecuteTypeSpecifyingExtension implements MethodTypeSpec
     {
         $args = $methodCall->getArgs();
 
-        if (0 === \count($args)) {
-            return null;
-        }
+       
 
         $stmtReflection = new PdoStatementReflection();
         $queryExpr = $stmtReflection->findPrepareQueryStringExpression($methodCall);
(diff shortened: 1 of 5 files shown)


vaielab:
I think it's also related to #770

staabm:
thanks for the PR. I do atm not have time to look into it. I will come back to it.

src/Extensions/PdoStatementExecuteTypeSpecifyingExtension.php
```diff
@@ -73,7 +73,32 @@ private function inferStatementType(MethodReflection $methodReflection, MethodCa
...
+                    $keyType = $scope->getType($bindArgs[0]->value);
+                    $constantStrings = $keyType->getConstantStrings();
+                    if ($keyType instanceof ConstantIntegerType) {
+                        $parameterKeys[] = $keyType;
+                        $parameterValues[] = $scope->getType($bindArgs[1]->value);
+                    } elseif ([] !== $constantStrings) {
+                        $parameterKeys[] = $constantStrings[0];
+                        $parameterValues[] = $scope->getType($bindArgs[1]->value);
+                    }
+                }
```
staabm, line 90: this code should be the same as in `PdoStatementExecuteMethodRule`.

we should not use `getConstantStrings` here

tests/default/data/pdo-stmt-execute.php
```diff
@@ -33,11 +34,12 @@ public function execute(PDO $pdo)
             assertType('array{email: string, 0: string, adaid: int<-32768, 32767>, 1: int<-32768, 32767>}', $row);
         }
 
-        $stmt = $pdo->prepare('SELECT email, adaid FROM ada WHERE adaid = ? and email = ?');
+        $stmt = $pdo->prepare('SELECT email, adaid FROM ada WHERE adaid = ? and email = ? ');
```
staabm: unrelates change?

```suggestion
        $stmt = $pdo->prepare('SELECT email, adaid FROM ada WHERE adaid = ? and email = ?');
```

staabm (changes requested):

src/Extensions/PdoStatementExecuteTypeSpecifyingExtension.php
vaielab: Thanks for taking the time to review this PR!

I can align this with the same pattern used in PdoStatementExecuteMethodRule.

However, when I do that, PHPStan reports:
"Doing instanceof PHPStan\Type\Constant\ConstantStringType is error-prone and deprecated. Use Type::getConstantStrings() instead."

It seems PdoStatementExecuteMethodRule does not trigger this because it’s currently ignored via phpstan-baseline.neon.

Would you prefer that I add a similar baseline exception here, or is there another approach you’d like me to follow?

staabm: yeah these errors - in this case - should be baselined.
I think it will be too hard to support union types for this API

staabm:
thank you!

PR #773, opened 2026-04-12
fix: support PDOStatement subclasses in PdoStatementExecuteMethodRule

vaielab:
Hello, me again! Here is another problem I found while using your extension. Hope this helps.

## Problem

`PdoStatementExecuteMethodRule` was skipping analysis when `execute()` is called on a subclass of `PDOStatement` that overrides the method. This is because the rule checked whether `execute()` was *declared* in `PDOStatement`:
 
```php
if (PDOStatement::class !== $methodReflection->getDeclaringClass()->getName()) {
  return [];
}
```
  
When a subclass overrides execute(), the declaring class is the subclass and not PDOStatement, so the rule bailed out early, silently missing placeholder errors.
 
## Fix
 
Replace the declaring-class check with a type check on the call's receiver:

```php
if (!(new \PHPStan\Type\ObjectType(PDOStatement::class))->isSuperTypeOf($scope->getType($methodCall->var))->yes()) {
  return []; 
}
```
 
This correctly handles PDOStatement itself and any subclass, regardless ofwhether execute() is inherited or overridden. 
 
## Tests
 
- Added MyPdoStatement fixture (extends PDOStatement, overrides execute())
- Added pdo-stmt-execute-subclassed.php test data with both error and no-error cases
- Added testSubclassedPdoStatement() to PdoStatementExecuteMethodRuleTest

### DIFF
--- a/src/Rules/PdoStatementExecuteMethodRule.php
+++ b/src/Rules/PdoStatementExecuteMethodRule.php
@@ -15,6 +15,7 @@ use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\ObjectType;
 use staabm\PHPStanDba\PdoReflection\PdoStatementReflection;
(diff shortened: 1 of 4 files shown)


staabm:
thank you!

PR #774, opened 2026-04-17
DoctrineKeyValueStyleRule: support named parameters

hemberger:
Since we configure which arguments to check by their position, this can break if the arguments are specified at the call site with named parameters (PHP 8.0+), which allows arguments to be passed out of order.

By mapping named parameter arguments to the actual index in the function interface, we can properly handle out of order calls.

### DIFF
--- a/src/Rules/DoctrineKeyValueStyleRule.php
+++ b/src/Rules/DoctrineKeyValueStyleRule.php
@@ -119,15 +119,22 @@ final class DoctrineKeyValueStyleRule implements Rule
             $methodReflection->getNamedArgumentsVariants(),
         );
 
-        $reorderedMethodCall = ArgumentsNormalizer::reorderMethodArguments(
-            $parametersAcceptor,
-            $callLike,
-        );
+        if ($callLike instanceof MethodCall) {
+            $reorderedCall = ArgumentsNormalizer::reorderMethodArguments(
+                $parametersAcceptor,
+                $callLike,
+            );
+        } else {
+            $reorderedCall = ArgumentsNormalizer::reorderNewArguments(
+                $parametersAcceptor,
+                $callLike,
+            );
+        }
 
-        if ($reorderedMethodCall === null) {
+        if ($reorderedCall === null) {
             return [];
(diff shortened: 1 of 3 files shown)


hemberger:
The PHP Linter test is failing because it runs in PHP 7.4. Since `tests/rules/data/doctrine-key-value-style-named-parameters.php` will only be able to be linted with PHP 8.0+, would it be okay to have parallel-lint exclude that file?

The other test failures appear to be unrelated to this PR.

I'm happy to make any changes you request. Thank you!

src/Rules/DoctrineKeyValueStyleRule.php
```diff
@@ -115,7 +115,30 @@ public function processNode(Node $callLike, Scope $scope): array
             return [];
         }
 
-        $tableExpr = $args[0]->value;
+        // Map function parameter names to parameter index
+        $params = $methodReflection->getVariants()[0]->getParameters();
```
staabm: can we utilize PHPStan native ArgumentsNormalizer?

staabm: usually PHPStan core would normalize arguments before passing them into rules.
I wonder why we need another normalization?

do we have similar problems in other rules?

tests/rules/data/doctrine-key-value-style-named-parameters.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
staabm:
```suggestion
<?php // lint >= 8.0
```

src/Rules/DoctrineKeyValueStyleRule.php
hemberger: I rewrote this with `ArgumentsNormalizer`, and it works as expected. Thanks!

> do we have similar problems in other rules?

I would expect that any Rule or RTE selecting arguments by index would have this same issue. It looks like many of them do, but I suspect that most people wouldn't be using named parameters with these kinds of functions (though they could in theory).

PR #776, opened 2026-05-17
Fix mixed row inference for CTE (WITH ...) queries

vaielab:
Fix row inference for queries starting with `WITH [RECURSIVE]`.

The issue caused iterated rows to be inferred as mixed instead of the expected typed array shape, because `getQueryType()` did not recognise CTE queries as SELECT-equivalent and bailed out before reaching the reflector.

Changes:
- add regression coverage for the CTE path (simple, multi, recursive, aggregation, SELECT *, FETCH_NUM)
- update QueryReflection::getQueryType() to skip past CTE definitions and return the inner statement keyword

Result:
foreach ($stmt as $row) now keeps the expected inferred array type for `WITH cte AS (...) SELECT ... FROM cte` queries.

### DIFF
--- a/.phpstan-dba-mysqli.cache
+++ b/.phpstan-dba-mysqli.cache
@@ -122,7 +122,7 @@
       array (
         5 => 
         array (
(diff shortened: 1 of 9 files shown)


staabm (changes requested):
it seems some of the added tests fail on some mariadb versions

PR #777, opened 2026-05-29
Preserve reflector-derived column types when AST narrowing yields ErrorType

ArtemGoutsoul:
# Preserve reflector-derived column types when AST narrowing yields `ErrorType`

## Issue

With `utilizeSqlAst(true)` + `defaultFetchMode(FETCH_TYPE_ASSOC)` on PHPStan **2.2+**, any `SELECT` that references columns through a table alias collapses every column's value type to `*ERROR*`:

```php
// service_group(service_group_id int unsigned, name varchar(45))

SELECT service_group_id, name FROM service_group
//  array{service_group_id: int<0, 4294967295>, name: string}        ✅

SELECT sg.service_group_id, sg.name FROM service_group sg
//  array{service_group_id: *ERROR*, name: *ERROR*}                  ❌
```

Worked on PHPStan 2.1; regressed in 2.2.

### Why

In `ParserInference::narrowResultType()`, per column:

1. `$valueType = $resultType->getOffsetValueType(ConstantIntegerType($i))` — but the `FETCH_TYPE_ASSOC` row shape from the reflector is string-keyed only. PHPStan 2.2 changed missing-offset reads to return `ErrorType` (previously benign).
2. `$queryScope->getType($expression)` is supposed to refine it. For unqualified columns it returns the real column type and overwrites the `ErrorType`. For alias-qualified columns (`t.col`) `resolveExpression()` has no branch for qualified identifiers and falls through to `MixedType`, so the guard `! $type instanceof MixedType` skips the assignment.
3. `$valueType` (still `ErrorType`) is then written over the correct reflector-derived type under the column-name key.

## Fix

After the QueryScope refinement step, skip the column when `$valueType` is still `ErrorType` — there is no information to add, so leave the reflector's type untouched.

```php
$type = $queryScope->getType($expression);
if (! $type instanceof MixedType) {
    $valueType = $type;
}

if ($valueType instanceof ErrorType) {
    continue;
}
```

Minimal, targeted: only changes behavior on the previously-broken path. A follow-up could teach `QueryScope::resolveExpression()` to resolve qualified identifiers and restore full AST narrowing (JOIN/WHERE nullability) for aliased columns, but that's a larger change.

## Test

Added `tests/default/ParserInferenceTest.php` — a plain PHPUnit test that drives `ParserInference::narrowResultType()` directly with a synthetic assoc-only `ConstantArrayType`, covering both unqualified and `t.col`-qualified queries. Fails without the patch (`*ERROR*` in the alias-qualified case), passes with it.

### DIFF
--- a/src/SqlAst/ParserInference.php
+++ b/src/SqlAst/ParserInference.php
@@ -8,7 +8,6 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
-use PHPStan\Type\ErrorType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use SqlFtw\Parser\Parser;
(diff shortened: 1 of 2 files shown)


src/SqlAst/ParserInference.php
```diff
@@ -179,6 +180,15 @@ public function narrowResultType(string $queryString, ConstantArrayType $resultT
...
             }
 
+            // PHPStan 2.2+ returns ErrorType when reading a missing offset (here
+            // the integer offset is missing on a string-keyed assoc row shape).
+            // If QueryScope couldn't refine the column either (e.g. alias-qualified
+            // references like `t.col`, which resolveExpression() returns MixedType
+            // for), there is nothing to narrow — keep the reflector-derived type.
+            if ($valueType instanceof ErrorType) {
+                continue;
+            }
```
staabm: could we check `$resultType->hasOffsetValueType()` before we call `$resultType->getOffsetValueType` instead?

ArtemGoutsoul: Good call — switched to a `hasOffsetValueType()` guard before the read. Same observable behavior, but no longer depends on the missing-offset return type. Matches the pattern used by the `setOffsetValueType()` calls below.

